PR #21, opened 2019-11-12
Polymorphic allocator

boeschf:
This is a proposal based on the standard libraries extension TS to support polymorphic allocators. This would allow the same message type for different architectures (host, device) and memory strategies (pools etc).


mbianco:
Temporarily closing it and reviving it later when needed

PR #25, opened 2019-11-22
Atlas_old

bettiolm:
Code for the integration with the Atlas library, based on a feature branch of Atlas which provides additional support for porting Atlas fields on the GPU and perform halo exchange in place. Main requirement for this feature is Atlas compiled with GridTools `data_store` dependency.


include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                 * number of levels is provided as well, defaul is 1 (2D case):
+                 * the assumption is that each 2D element is a column, with 'levels' vertical elements.
+                 * WARN: a second index, and therefore a more complex iteration space,
+                 * is needed to handle correctly multiple vertical layers*/
+                class iteration_space {
+
+                    private:
+
+                        int m_partition;
+                        std::vector<index_type, gridtools::ghex::allocator::unified_memory_allocator<index_type>> m_local_index;
```
mbianco: The only supported allocator is the UVM allocator? Could this be a trait so we can reuse the code with another partitioner? This I think would come up automatically when working with parmetis (or whatever we choose). We also should think about AMD which does not have UVM, ASAICT

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                        std::size_t m_levels;
+
+                    public:
+
+                        // ctors
+                        iteration_space() noexcept = default;
+                        iteration_space(const int partition,
+                                const std::vector<index_type, gridtools::ghex::allocator::unified_memory_allocator<index_type>>& local_index,
+                                const std::size_t levels = 1) noexcept :
```
mbianco: I think this can throw

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                            m_partition{partition},
+                            m_local_index{std::move(local_index)},
+                            m_levels{levels} {}
+                        // less safe but maybe preferable
+                        // template<typename V>
+                        // iteration_space(const int partition, V&& remote_index) noexcept :
+                        //     m_partition{partition},
+                        //     m_remote_index{std::forward<V>(remote_index)} {}
+                        iteration_space(const int partition, const index_type first, const index_type last, const std::size_t levels = 1) noexcept :
```
mbianco: This one also can throw

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                            m_local_index{},
+                            m_levels{levels} {
+                            m_local_index.resize(static_cast<std::size_t>(last - first + 1));
+                            for (index_type idx = first; idx <= last; ++idx) {
+                                m_local_index[static_cast<std::size_t>(idx)] = idx;
+                            }
+                        }
+
+                        // member functions
+                        int partition() const noexcept { return m_partition; }
```
mbianco: In this implementation we have a single subdomain per rank, right? Is this why this returns always the same value?

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                        // std::vector<index_type, gridtools::allocator::unified_memory_allocator<index_type>>& local_index() noexcept { return m_local_index; }
```
mbianco:
```suggestion
```

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                        const std::vector<index_type, gridtools::ghex::allocator::unified_memory_allocator<index_type>>& local_index() const noexcept { return m_local_index; }
```
mbianco: Calling this `local_index` makes me think it's a single value, while the use should be `obj.local_index[i]`. Should we make clear from the name that this returns a container?

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+ */
+#ifndef INCLUDED_GHEX_UNSTRUCTURED_PATTERN_HPP
+#define INCLUDED_GHEX_UNSTRUCTURED_PATTERN_HPP
+
+#include <vector>
+#include <map>
+#include <cassert>
+#include <numeric>
+#include <cstring>
+
```
mbianco:
```suggestion
#include <iosfwd>

```
?

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                    // bool operator < (const extended_domain_id_type& other) const noexcept {
+                    //     return (id < other.id ? true : (id == other.id ? (tag < other.tag) : false));
+                    // }
+                    /** @brief unique ordering given by address and tag*/
+                    bool operator < (const extended_domain_id_type& other) const noexcept {
+                        return (address < other.address ? true : (address == other.address ? (tag < other.tag) : false));
+                    }
+
+                    // print
+                    // /** @brief print*/
```
mbianco: Clean up ?

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                        }
+
+                };
+
+                using iteration_space_pair = iteration_space;
+                using index_container_type = std::vector<iteration_space_pair>;
+
+                /** @brief extended domain id, including rank and tag information
+                 * WARN: domain id temporarily set equal to rank, differently from structured pattern case*/
+                struct extended_domain_id_type {
```
mbianco: Maybe we should put these classes outside the scope of the class, maybe in a namespace, because reading the code is not easy now.

include/ghex/unstructured/pattern.hpp
```diff
@@ -0,0 +1,346 @@
...
+                map_type m_recv_map;
+                pattern_container_type* m_container;
+
+            public:
+
+                // ctors
+                pattern(communicator_type& comm, const iteration_space_pair& domain, const extended_domain_id_type& id) :
+                    m_comm(comm),
+                    m_domain(domain),
+                    m_id(id) {}
```
mbianco: Where are the other members set? Expecially `m_container`, should it be set to nullptr?

bettiolm:
I will rename thi PR into Atlas_old and create a new one from a fork, otherwise I will not be able to add further commits here. I will keep this one as reference, at least until we merge.

benchmarks/atlas_halo_exchange_nodecolumns.cpp
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * Please, refer to the LICENSE file in the root directory.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <vector>
+#include <sys/time.h>
+
+//#include "atlas/parallel/mpi/mpi.h"
+#include "atlas/library/Library.h"
```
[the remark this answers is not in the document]
bettiolm: Fixed

benchmarks/atlas_halo_exchange_nodecolumns.cpp
```diff
@@ -0,0 +1,194 @@
...
+    // Instantate halo generator
+    gridtools::ghex::atlas_halo_generator<int> hg{rank, size};
+
+    // Make patterns
+    auto patterns = gridtools::ghex::make_pattern<gridtools::ghex::unstructured::grid>(mpi_comm, hg, local_domains);
+
+    // Istantiate communication object
+    using communication_object_t = gridtools::ghex::communication_object<decltype(patterns)::value_type, gridtools::ghex::cpu>;
+    std::vector<communication_object_t> cos;
+    for (const auto& p : patterns) {
```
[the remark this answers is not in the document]
bettiolm: This is due to the fact that here I am using the simple communicator. I changed the code in order to use `communicator_object_2` everywhere, except from the first test of the Atlas series.

include/ghex/allocator/unified_memory_allocator.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace gridtools {
+    namespace ghex {
+        namespace allocator {
+
+            // WARN: some member types and member functions provided in std::allocator<T> are still not provided here;
+            // WARN: fails to compile is inheritance is used (potential issues on vector copy constructor when using different allocators)
+            template <class T>
+            struct unified_memory_allocator {
+
```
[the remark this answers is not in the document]
bettiolm: Added, thanks.

include/ghex/transport_layer/mpi/setup.hpp
```diff
@@ -135,8 +136,41 @@ namespace gridtools{
...
                         &h.get()));
                     return {std::move(res), std::move(h)};
-                } 
-
+                }
+                
+                /** @brief just a helper function using custom types to be used when send/recv counts can be deduced*/
+                template<typename T>
+                void allToAll(const std::vector<T>& send_buf, std::vector<T>& recv_buf) const
+                {
```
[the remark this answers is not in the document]
bettiolm: Done

include/ghex/unstructured/pattern.hpp
bettiolm: Not sure.. Why would I need it here?

include/ghex/unstructured/pattern.hpp
bettiolm: Totally agree. My idea would be to keep it like this for the scope of this PR, and then do some refactoring for the general unstructured support (parmetis...).

include/ghex/unstructured/pattern.hpp
bettiolm: Why only this one and not the following?

include/ghex/unstructured/pattern.hpp
mbianco: you have a std::ostream reference in line 151, with this you ensure the declaration of that is available, and you would not need to include iostream if you do not need to use. Not only, this would make the inclusion of iostream not be forced to happen before this file

include/ghex/unstructured/pattern.hpp
mbianco: I think the move constructor of the vector does not throw. If it does, then also the next one can throw

include/ghex/unstructured/pattern.hpp
bettiolm: Yep, true. I pushed it to the new PR.

include/ghex/unstructured/pattern.hpp
bettiolm: Ok, done.

include/ghex/unstructured/pattern.hpp
bettiolm: Done

include/ghex/unstructured/pattern.hpp
bettiolm: Yes, this implementation assumes only one domain per rank, and I take advantage of it. Partition and rank are basically mapped 1:1. Of course this is going to change in a more general implementation.

include/ghex/unstructured/pattern.hpp
bettiolm: Ok

include/ghex/unstructured/pattern.hpp
bettiolm: True. I kept this from Atlas convention, in which all objects like, e,g, partition, remote_index ... are left singular since they are actually `fields`. I can change this everywhere, even if I think I am going to refactor it a little bit on a following PR for a more general approach.

include/ghex/unstructured/pattern.hpp
bettiolm: Yep, I agree. Only thing is that I think we should be consistent in `structured/pattern.hpp` as well, so if you say I can refactor that one as well.

include/ghex/unstructured/pattern.hpp
bettiolm: Done

include/ghex/unstructured/pattern.hpp
bettiolm: Yep, right. Done.

PR #38, opened 2020-01-06
PMIx db interface

angainor:
and modified transport benchmarks


include/ghex/transport_layer/ucx/context.hpp
```diff
@@ -233,14 +233,19 @@ namespace gridtools {
...
-                static std::unique_ptr<context<ucx_tag, ThreadPrimitives>> create(int num_threads, MPI_Comm mpi_comm = MPI_COMM_NULL)
+                static std::unique_ptr<context<ucx_tag, ThreadPrimitives>> create(int num_threads, MPI_Comm mpi_comm)
                 {
 #ifdef GHEX_USE_PMI
                     return std::make_unique<context<ucx_tag,ThreadPrimitives>>(num_threads, mpi_comm, ucx::address_db_pmi{mpi_comm});
 #else
                     return std::make_unique<context<ucx_tag,ThreadPrimitives>>(num_threads, mpi_comm, ucx::address_db_mpi{mpi_comm});
 #endif
                 }
+                template<typename DB>
```
angainor: @boeschf Changed, as we discussed in the original PR https://github.com/boeschf/GHEX/pull/3#discussion_r359734426. Please have a look if this is OK.

benchmarks/transport/CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ if (GHEX_USE_UCP)
         if (GHEX_USE_PMIX)
             target_compile_definitions(${_t} PRIVATE USE_PMIX)
         endif()
+        if (GHEX_USE_UCP)
```
angainor: @boeschf @mbianco We have `GHEX_USE_UCP` and `GHEX_USE_PMIX` defined in `cmake`, while I was using `USE_UCX` and `USE_PMIX` defines in my code. I guess the `GHEX_USE_*` is your preferred naming strategy? If that is the case, then I will simply move all my defines to that scheme.

boeschf: I think `GHEX_` prefix is better to avoid any clashes

include/ghex/transport_layer/ucx/context.hpp
boeschf: does it work above when passing an MPI comm and creating a pmi database (when `GHEX_USE_PMI` is enabled)?

angainor: @boeschf yes, but at this moment the mpi comm is null in this code path - because this has to be called before MPI_Init. So the code will work, but for the barrier to work the comm needs to be updated later to a correct value. Do we want a set function for the mpi comm? 

Right now I've modified the benchmarks to use the other code path.

boeschf: I think we should not use it at all, to be safe (I don't like MPI == null). The second function which takes a database may be modified to take only a db and get the mpi-comm from it (that would eliminate a potential user error) - it would require to update the database classes, though. There could even be a default database argument (the mpi db). Then the first function would not be needed anymore.

angainor: I don't think we can avoid `MPI_COMM_NULL`. I need to construct the `transport_context` before  `MPI_Init` is called. And in your implementation `transport_context` takes the `parallel_context` as construction argument, which requires mpi comm to be passed. This dependency has to be broken, and it can be done using `MPI_COMM_NULL`, or dropping the `parallel_context` from `transport_context` constructor.

include/ghex/transport_layer/ucx/context.hpp
```diff
@@ -171,73 +178,63 @@ namespace gridtools {
...
-                    else
-                        throw std::runtime_error("could not connect to endpoint");
+                    auto addr = m_context->m_db.find(rank);
+                    auto p = m_endpoint_cache.insert(std::make_pair(rank, endpoint_t{rank, m_worker.get(), addr}));
+                    return p.first->second;
                 }
 
             } // namespace ucx
-            
-            template<class ThreadPrimitives>
```
angainor: @boeschf I moved the context factory to `transport_layer/context.cpp` - I don't see a need to specialize here.

cmake/FindPMIx.cmake
```diff
@@ -3,13 +3,14 @@ pkg_check_modules(PC_PMIX QUIET pmix)
...
 find_path(PMIX_INCLUDE_DIR pmix.h
     HINTS
-        ${PMIX_ROOT}  ENV PMIX_ROOT
-        ${PMIX_DIR}   ENV PMIX_DIR
+    ${PMIX_ROOT}  ENV PMIX_ROOT
+    ${PMIX_DIR}   ENV PMIX_DIR
     PATH_SUFFIXES include)
 
-find_library(PMIX_LIBRARY NAMES pmix
+find_library(PMIX_LIBRARY HINT ${PMIX_DIR} NAMES pmix
```
angainor: For symmetry with UCX - this actually lets cmake find the library provided by a module, if the module specifies `PMIX_DIR`

include/ghex/transport_layer/context.hpp
```diff
@@ -139,39 +31,34 @@ namespace gridtools {
...
-                /*const */thread_primitives_type& thread_primitives() /*const*/ noexcept 
-                {
-                    return m_parallel_context.thread_primitives(); 
+                    return m_thread_primitives;
                 }
 
                 mpi::setup_communicator get_setup_communicator()
                 {
-                    return mpi::setup_communicator(m_parallel_context.world());
+                    // TODO: what about this?
```
angainor: @boeschf This is hanging: since we do not have the mpi comm here anymore, the setup communicator should be moved elsewhere. Do you need it here?

include/ghex/transport_layer/context.hpp
```diff
@@ -191,19 +78,28 @@ namespace gridtools {
...
-                    return m_parallel_context.m_thread_primitives.get_token();
+                    return m_thread_primitives.get_token();
                 }
-                
+
                 // thread-safe
                 void barrier(thread_token& t) /*const*/
                 {
-                    m_parallel_context.barrier(t);
+                    m_thread_primitives.barrier(t);
```
angainor: @boeschf The barrier has been moved to the transport context, as we discussed. I still need to implement a trivial barrier for UCX transport.

include/ghex/transport_layer/context.hpp
```diff
@@ -191,19 +78,28 @@ namespace gridtools {
...
+                    m_transport_context.barrier();
+                    m_thread_primitives.barrier(t);
                 }
 
             };
-            
+
             template<class TransportTag, class ThreadPrimitives>
-            struct context_factory;
+            struct context_factory
```
angainor: @boeschf and here is the new generic context factory.

include/ghex/transport_layer/ucx/context.hpp
boeschf: There are differences in how one constructs the context (ucx vs. mpi vs. libfabric), so I would leave it where it is.

cmake/FindPMIx.cmake
boeschf: sure

include/ghex/transport_layer/ucx/context.hpp
boeschf: For example: the database is not an argument that would be needed for mpi

angainor: OK

include/ghex/transport_layer/context.hpp
boeschf: Yes I do need it, this should maybe ask the transport context for the mpi-comm?

angainor: @boeschf the transport context doesn't have the MPI comm :) at least not the UCX context

include/ghex/transport_layer/ucx/context.hpp
angainor: @boeschf moved it back as it was

include/ghex/transport_layer/context.hpp
boeschf: there needs to be a mapping to an mpi comm at some point somewhere - that's a requirement we have for the high level API (to be able to figure out halos, to be able to do collectives, etc.). We need to rethink this initialization business

include/ghex/transport_layer/context.hpp
```diff
@@ -139,39 +31,35 @@ namespace gridtools {
...
-                /*const */thread_primitives_type& thread_primitives() /*const*/ noexcept 
-                {
-                    return m_parallel_context.thread_primitives(); 
+                    return m_thread_primitives;
                 }
 
                 mpi::setup_communicator get_setup_communicator()
                 {
-                    return mpi::setup_communicator(m_parallel_context.world());
+                    return mpi::setup_communicator(m_mpi_comm);
```
angainor: @boeschf Have a look at this one. Is this doing what it's supposed to now?

boeschf: this looks good now

include/ghex/transport_layer/ucx/context.hpp
```diff
@@ -171,73 +178,74 @@ namespace gridtools {
...
             template<class ThreadPrimitives>
             struct context_factory<ucx_tag, ThreadPrimitives>
             {
-                static std::unique_ptr<context<ucx_tag, ThreadPrimitives>> create(int num_threads, MPI_Comm mpi_comm)
+                template<typename DB>
+                static std::unique_ptr<context<ucx_tag, ThreadPrimitives>> create(int num_threads, DB&& db)
                 {
-                    return std::make_unique<context<ucx_tag,ThreadPrimitives>>(num_threads, mpi_comm, ucx::address_db_mpi{mpi_comm});
+                    return std::make_unique<context<ucx_tag,ThreadPrimitives>>(num_threads, db.m_mpi_comm, std::forward<DB>(db));
                 }
```
boeschf: since all the databases take a mpi_comm now, couldn't you construct the dbs inside this create function (and have the ifdefs there)?

include/ghex/transport_layer/ucx/address_db_mpi.hpp
```diff
@@ -50,15 +50,14 @@ namespace gridtools {
...
                     {
                         auto it = m_address_map.find(k);
                         if (it != m_address_map.end())
                         {
-                            return &(it->second);
+                            return it->second;
                         }
-                        else
-                            return nullptr;
+                        throw std::string("Cound not find peer address in the MPI address xdatabase.");
```
boeschf: please throw a runtime_error and not a simple string

include/ghex/transport_layer/ucx/context.hpp
```diff
@@ -171,73 +178,78 @@ namespace gridtools {
...
-                    
+
                 rank_type rank() const { return m_db.rank(); }
                 rank_type size() const { return m_db.size(); }
                 ucp_context_h get() const noexcept { return m_context.m_context; }
 
+                void barrier()
+                {
+                }
```
boeschf: can you come up with an implementation of the barrier? do we need it for the benchmarks/tests?

include/ghex/transport_layer/util/pmi/pmix/pmi.hpp
```diff
@@ -23,133 +23,160 @@ namespace gridtools
...
-		    PMIX_VALUE_RELEASE(pvalue);
-		    PMIX_PROC_DESTRUCT(&proc);
-
-		    LOG("PMIx_get %s returned %zi bytes", key.c_str(), data.size());
-		    
-		    return  data;
-		}
-	    };
-	} // namespace tl
+        namespace tl
```
boeschf: this should probably be in its own `namespace` (since it is not in the folder `transport_layer`)

include/ghex/transport_layer/util/pmi/pmix/pmi.hpp
```diff
@@ -23,133 +23,160 @@ namespace gridtools
...
+
+                ~pmi()
+                {
+                    int rc;
+                    if (PMIX_SUCCESS != (rc = PMIx_Finalize(NULL, 0))) {
+                        WARN("Client ns %s rank %d:PMIx_Finalize failed: %d\n", myproc.nspace, myproc.rank, rc);
+                    } else {
+                        if(myproc.rank == 0) LOG("PMIx finalized");
+                    }
+                }
```
boeschf: this class manages a resource (`pmix`) and thus it should either be **_non-copyable_** (move-only) or hold the resource in a `shared_ptr`

include/ghex/transport_layer/ucx/address_db_pmi.hpp
```diff
@@ -0,0 +1,105 @@
...
+                        std::vector<unsigned char> data(addr.data(), addr.data()+addr.size());
+                        pmi_impl_static.set("ghex-rank-address", data);
+
+                        // TODO: we have to call an explicit PMIx Fence due to
+                        // https://github.com/open-mpi/ompi/issues/6982
+                        pmi_impl_static.exchange();
+                    }
+                };
+
+                PmiType address_db_pmi::pmi_impl_static;
```
boeschf: is this a problem when we deal with multiple translation units (library scenario)?

include/ghex/transport_layer/ucx/context.hpp
angainor: done

include/ghex/transport_layer/util/pmi/pmix/pmi.hpp
angainor: changed to `gridtools::ghex::tl::pmi`

include/ghex/transport_layer/ucx/address_db_mpi.hpp
angainor: done

include/ghex/transport_layer/ucx/address_db_pmi.hpp
angainor: I have not experienced any.

boeschf: since it is marked static, each translation unit would have it's own copy

angainor: the static variable is not needed anymore, since we have to run a data exchange anyway. it was needed when I had to initialize PMIx before we initialized MPI. I removed it.

include/ghex/transport_layer/util/pmi/pmix/pmi.hpp
angainor: Please have a look if this is what you had in mind. I only check for moved bit in the destructor though. Seems to me like it is not needed elsewhere, but maybe I'm wrong.

benchmarks/transport/CMakeLists.txt
angainor: My approach here would be to collect such things in the `config.h` file and not the commandline. I had a PR for that. Not sure if you guys will want that or not.

include/ghex/transport_layer/util/pmi/pmix/pmi.hpp
```diff
@@ -17,139 +17,168 @@
...
+                    }
+
+                    void set(const std::string key, const std::vector<unsigned char> data)
+                    {
+                        int rc;
+                        pmix_value_t value;
+
+                        PMIX_VALUE_CONSTRUCT(&value);
+                        value.type = PMIX_BYTE_OBJECT;
+                        value.data.bo.bytes = (char*)((void*)data.data());
```
boeschf: why cast twice? why not `reinterpret_cast<char*>(data.data())`

boeschf: maybe you need a const_cast as well here

include/ghex/transport_layer/util/pmi/pmix/pmi.hpp
```diff
@@ -17,139 +17,168 @@
...
+                            throw std::runtime_error("PMIx_Init failed with code " + std::to_string(rc));
+                        }
+                        if(myproc.rank == 0) LOG("%d PMIx initialized", myproc.rank);
+
+                        /* job-related info is found in our nspace, assigned to the
+                         * wildcard rank as it doesn't relate to a specific rank. Setup
+                         * a name to retrieve such values */
+                        PMIX_PROC_CONSTRUCT(&allproc);
+                        // (void)strncpy(allproc.nspace, myproc.nspace, PMIX_MAX_NSLEN);
+                        (void)memcpy(allproc.nspace, myproc.nspace, PMIX_MAX_NSLEN);
```
boeschf: can you use std::memcpy here?

angainor: done

PR #39, opened 2020-01-06
Cmake change

angainor:
IMO, this config option is not needed. This is a runtime setting, not related to compilation in any way. Also, with new OpenMPI versions `-mca pml ucx` is the default on most systems. If it is not (Cray?), we should probably configure this on the OpenMPI module level.

In either case, this would only work for OpenMPI.


boeschf:
Agreed, this can go away - I will merge if nobody else complains

PR #41, opened 2020-01-17
Uncrustify

boeschf:


cmake/FindClangTools.cmake
```diff
@@ -0,0 +1,45 @@
...
+#                       locations to find clang-tidy and clang-format
+#
+# This module defines:
+#  CLANG_TIDY_BIN       The  path to the clang tidy binary
+#  CLANG_TIDY_FOUND     Whether clang tidy was found
+#  CLANG_FORMAT_BIN     The path to the clang format binary 
+#  CLANG_FORMAT_FOUND   Whether clang format was found
+#
+
+find_program(CLANG_TIDY_BIN
```
mbianco: Do we need this now?

angainor: I think this is the best way to do this. We push scripts, and the actual formatting is moved to another PR.

mbianco: We should then revert the other files and make a small PR

PR #43, opened 2020-01-21
Pthread recursive

boeschf:


include/ghex/threads/omp/primitives.hpp
```diff
@@ -73,6 +74,8 @@ namespace gridtools {
 
                     using mutex_type = ::gridtools::ghex::threads::mutex::pthread_spin::mutex;
                     using lock_type  = ::gridtools::ghex::threads::mutex::pthread_spin::lock_guard;
+                    //using mutex_type = ::gridtools::ghex::threads::mutex::pthread::mutex;
```
mbianco: Are these options to try? Should we guard them with macros?

boeschf: these are options which we may use for different network architectures. it turned out that for infiniband the spinlock was the most efficient. Would probably be a good idea to add macros, but we need a place where we configure the library.

mbianco: like the `config.h` mentioned by Marcin some time ago?

boeschf: yes, that would be the place I guess

mbianco:
Can this be merged?

PR #44, opened 2020-01-22
MPI API update

boeschf:


include/ghex/transport_layer/mpi/communicator.hpp
```diff
@@ -28,162 +30,226 @@ namespace gridtools {
...
+
+                    unsigned progress() { return m_state->progress(); }
+
+                    void barrier() {
+                        if (auto token_ptr = m_state->m_token_ptr) {
+                            auto& tp = *(m_shared_state->m_thread_primitives);
+                            auto& token = *token_ptr;
+                            tp.barrier(token);
+                            tp.single(token, [this]() { MPI_Barrier(m_shared_state->m_comm); } );
+                            tp.barrier(token);
```
mbianco: Why do you need a second thread barrier?

include/ghex/transport_layer/mpi/communicator.hpp
```diff
@@ -28,162 +30,226 @@ namespace gridtools {
...
-                      * @param tag message tag
-                      * @param args additional arguments to be passed to new container of type Message at construction 
-                      * @return optional which may hold a future< std::tuple<Message,rank_type,tag_type> > */
-                    template<typename Message, typename... Args>
-                    [[nodiscard]] auto recv_any_source_any_tag(Args&& ...args) const
+                        GHEX_CHECK_CALLBACK_F(message_type,rank_type,tag_type) 
+                        return recv(message_type{shared_msg.m_message}, src, tag, std::forward<CallBack>(callback));
+                    }
+
+                    template<typename Message, typename CallBack>
```
mbianco: We need documentation (doxygen style) on these functions

include/ghex/transport_layer/mpi/communicator.hpp
```diff
@@ -28,162 +30,226 @@ namespace gridtools {
...
                     {
-                        return recv_any<Message>(MPI_ANY_SOURCE, MPI_ANY_TAG, std::forward<Args>(args)...);
+                        GHEX_CHECK_CALLBACK_F(message_type,rank_type,tag_type) 
+                        using V = typename std::remove_reference_t<Message>::value_type;
+                        return recv(message_type{ref_message<V>{msg.data(),msg.size()}}, src, tag, std::forward<CallBack>(callback));
                     }
 
-                private: // implementation
+                    template<typename Message, typename CallBack>
+                    rvalue_func<Message&&> recv(Message&& msg, rank_type src, tag_type tag, CallBack&& callback, std::true_type)
```
mbianco: Wouldn't it make more sense to just call the function with a different type instead of having this extra argument?

include/ghex/transport_layer/mpi/communicator_base.hpp
```diff
@@ -11,10 +11,11 @@
 #ifndef INCLUDED_GHEX_TL_MPI_COMMUNICATOR_BASE_HPP
 #define INCLUDED_GHEX_TL_MPI_COMMUNICATOR_BASE_HPP
 
+#include "./error.hpp"
 #include <memory>
-#include <omp.h>
-#include "./request.hpp"
-#include "./status.hpp"
+//#include <omp.h>
```
mbianco:
```suggestion
```

include/ghex/transport_layer/mpi/communicator_base.hpp
```diff
@@ -11,10 +11,11 @@
...
+//#include "./request.hpp"
```
mbianco:
```suggestion
```

include/ghex/transport_layer/mpi/communicator_base.hpp
```diff
@@ -11,10 +11,11 @@
...
+//#include "./status.hpp"
```
mbianco:
```suggestion
```

include/ghex/transport_layer/mpi/communicator_state.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+namespace gridtools {
+
+    namespace ghex {
+
+        namespace tl {
+
+            namespace mpi {
+
+                template<typename ThreadPrimitives>
```
mbianco: Documentation of the purpose of these classes.

include/ghex/transport_layer/mpi/setup.hpp
```diff
@@ -12,6 +12,8 @@
 #define INCLUDED_GHEX_TL_MPI_SETUP_HPP
 
 #include "./communicator_base.hpp"
+//#include "./request.hpp"
```
mbianco:
```suggestion
```

include/ghex/transport_layer/mpi/communicator.hpp
boeschf: this extra argument was a bug - I corrected it

include/ghex/transport_layer/mpi/communicator.hpp
boeschf: I am not sure if we need it here. We certainly need it for the ucx communicator. Maybe @angainor can comment as well.

angainor: In UCX we need both to make sure that only a single thread progresses the communication. Here we probably don't, but it's a question of which one do you want to keep: the first one, or the second one. I'd probably keep the second one only: let the first thread start the global barrier immediately, without having to wait for all the others. Could be more efficient, but then again we are not targeting efficiency here.

tests/transport/test_cancel.cpp
```diff
@@ -0,0 +1,100 @@
...
+    else {
+        using comm_t = typename Context::communicator_type;
+        using msg_t = typename comm_t::message_type;
+        using rank_t = typename comm_t::rank_type;
+        using tag_t = typename comm_t::tag_type;
+
+        int counter = 0;
+        auto req = comm.recv(msg, 0, 42, [&counter](msg_t, rank_t, tag_t){ ++counter; });
+        bool ok = req.cancel();
+        for (int i=0; i<100; ++i)
```
mbianco: Why `100`?

tests/transport/test_cancel.cpp
```diff
@@ -0,0 +1,100 @@
...
+            comm.progress();
+        EXPECT_TRUE(counter == 0);
+        comm.recv(msg, 0, 42+42).wait();
+        for (int i=0; i<100; ++i)
```
mbianco: Why `100`?

tests/transport/test_cancel.cpp
boeschf: a significant amount of times - 100 is an arbitrary value. I could add a timer instead, would that be better?

mbianco: What would be a reasonable amount of time? My problem here is: can this test fails just because the time is too short?

boeschf: It is more the other way round: you could miss a fail when not waiting long enough - a false positive.

include/ghex/structured/pattern.hpp
```diff
@@ -12,7 +12,7 @@
 #define INCLUDED_GHEX_STRUCTURED_PATTERN_HPP
 
 #include "./grid.hpp"
-#include "../transport_layer/communicator.hpp"
+#include "../transport_layer/tags.hpp"
```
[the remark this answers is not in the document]
boeschf: No, I will remove it

tests/transport/test_context.cpp
```diff
@@ -0,0 +1,196 @@
...
+using transport = gridtools::ghex::tl::mpi_tag;
+#endif
+
+using threading_std = gridtools::ghex::threads::std_thread::primitives;
+using threading_atc = gridtools::ghex::threads::atomic::primitives;
+using threading = threading_std;
+//using threading = threading_atc;
+using context_type = gridtools::ghex::tl::context<transport, threading>;
+using comm_type = typename context_type::communicator_type;
+using msg_type = typename comm_type::message_type;
```
mbianco: It seems these types are defined here, but then every test instantiate it's own context type. Would it make more sense to make every test to instantiate their own types?

tests/transport/test_cancel.cpp
angainor: @boeschf Why do you need the loop here anyway? the canceled recv expects tag 42, which you don't send. So counter will always be 0, and although canceling of a UCX request requires a call to progress, we return true and don't really wait for it to happen.

tests/transport/test_cancel.cpp
angainor: @boeschf same here - do you need this loop? `wait` progresses already.

tests/transport/test_barrier_mt.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+TEST(transport, barrier_mt_std) {
+    auto context_ptr = gridtools::ghex::tl::context_factory<transport,threading>::create(4, MPI_COMM_WORLD);
+    auto& context = *context_ptr;
+    
+    auto func = [&context]() {
+        auto token = context.get_token();
+        auto comm = context.get_communicator(token);
+
+        for(int i=0; i<50; i++)  {
```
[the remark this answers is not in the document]
boeschf: it is already quite slow. I'll try to restructure the test

tests/transport/test_cancel.cpp
boeschf: `wait` does not progress when the backend is mpi

tests/transport/test_cancel.cpp
boeschf: I guess we could return some more status information from `progress()` -  not a single number but number of cancels etc. what do you think? this way we could also eliminate the sometimes spurious results from ucx when we detect ucx-internal messages which we didn't send.

mbianco: I do not know if this can help, but returning something related to the internal state of the queues could be used there. Of course, if the queues are empty this does not mean that there are not messages still in flight somewhere

angainor: @boeschf we don't have control over what the ucx worker progress returns, and we don't know what has finished during a progress call. I think it will be problematic to figure this out in general for shared workers. 

We also don't control whether the request is canceled or completed - what we know is that after a call to `cancel` it will eventually be either completed, or canceled, and a callback will be called with the corresponding state (in UCX). If you want to make sure that it has actually been canceled, then we could set some flag in the communicator recv callback.

tests/transport/test_cancel.cpp
angainor: @boeschf maybe it should to be consistent with the UCX backend?

boeschf:
added a more detailed return type for `progress()`, please check the `test_cancel.cpp`.

tests/transport/test_cancel.cpp
boeschf: Maybe it should, but I am not sure this is necessary. I believe, requiring the user to call progress explicitly if they are using callback-based communication is fine. If this was the requirement (as documented in the doxygen strings), then the backends behave consistently.

tests/transport/test_barrier_mt.cpp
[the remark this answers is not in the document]
boeschf: It is slow for the busy-wait based mutexes on my machine (since I am using more threads as I have physical cores - take into account, that mpi itself is spawning several processes itself). The only variant that is performant for me is `std::mutex` (the first test). I guess we have the following options
- just wait forever on laptops
- test only 50 iterations
- apply some cmake logic to configure the tests

tests/transport/test_context.cpp
boeschf: I changed the test: is this what you meant?

tests/transport/test_cancel.cpp
boeschf: @angainor I believe by tracking the progress (number of called callbacks/number of cancels) ourselves we are safe. take a look at the suggested implementation in the new commits and let me know what you think.

tests/transport/test_barrier_mt.cpp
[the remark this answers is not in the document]
mbianco: For now I guess it's ok, if there is a bug in a barrier we should be able to catch it sooner or later...

include/ghex/transport_layer/callback_utils.hpp
```diff
@@ -255,7 +255,7 @@ namespace gridtools {
                     }
                 };
 
-                progress_status operator+(progress_status a, progress_status b) { return a+=b; }
+                static progress_status operator+(progress_status a, progress_status b) { return a+=b; }
```
mbianco: Why isn't 
```suggestion
                static progress_status operator+(progress_status a, progress_status b) { return a+b; }
```

tests/transport/test_barrier.cpp
```diff
@@ -0,0 +1,35 @@
...
+    auto comm = context.get_communicator(token);
+    int rank = context.rank();
+    gridtools::ghex::timer timer;
+
+    timer.tic();
+    for(int i=0; i<100000; i++)  {
+        comm.barrier();
+    }
+    const auto t = timer.stoc();
+    if(rank==0)
```
mbianco: not really sure you should print the times here...

PR #49, opened 2020-01-24
change UCX initialization to single-threaded

angainor:
This is OK according to the developers, since we anyways lock UCX from outside.

https://github.com/openucx/ucx/issues/4609

Could provide some performance benefits - will check.

### DIFF
--- a/include/ghex/transport_layer/ucx/context.hpp
+++ b/include/ghex/transport_layer/ucx/context.hpp
@@ -155,15 +155,14 @@ namespace gridtools {
                     // check the actual parameters
                     ucp_context_attr_t attr;
                     attr.field_mask =
-                        UCP_ATTR_FIELD_REQUEST_SIZE | // internal request size
-                        UCP_ATTR_FIELD_THREAD_MODE;   // thread safety
+                        UCP_ATTR_FIELD_REQUEST_SIZE ; // internal request size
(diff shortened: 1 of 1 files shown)


include/ghex/transport_layer/ucx/context.hpp
```diff
@@ -155,15 +155,14 @@ namespace gridtools {
                     // check the actual parameters
                     ucp_context_attr_t attr;
                     attr.field_mask =
-                        UCP_ATTR_FIELD_REQUEST_SIZE | // internal request size
-                        UCP_ATTR_FIELD_THREAD_MODE;   // thread safety
+                        UCP_ATTR_FIELD_REQUEST_SIZE ; // internal request size
                     ucp_context_query(m_context.m_context, &attr);
                     m_req_size = attr.request_size;
-                    if (attr.thread_mode != UCS_THREAD_MODE_MULTI)
-                        throw std::runtime_error("ucx cannot be used with multi-threaded context");
```
boeschf: why are you not checking for multi-threaded ucx context anymore here?

angainor: This checks whether UCX supports multithreading. We don't need it to support multithreading anymore - we can use UCX compiled in single-threaded mode.

PR #52, opened 2020-01-28
fix ep cleanup

angainor:


include/ghex/transport_layer/ucx/endpoint.hpp
```diff
@@ -65,8 +69,16 @@ namespace gridtools {
 
                     void destroy()
                     {
-                        if (!m_moved)
-                            ucp_ep_close_nb(m_ep, UCP_EP_CLOSE_MODE_FLUSH);
+                        if (!m_moved){
+                            ucs_status_ptr_t ret = ucp_ep_close_nb(m_ep, UCP_EP_CLOSE_MODE_FLUSH);
+                            if(UCS_OK == reinterpret_cast<std::uintptr_t>(ret)) return;
+                            if(UCS_PTR_IS_ERR(ret)) return;
```
boeschf: should we print a warning here?

angainor: About the error? maybe. but we don't have yet a proper warning/logging facility

PR #76, opened 2020-03-10
node local ranks

boeschf:


include/ghex/transport_layer/mpi/hwtopo.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <unordered_set>
+#include "./error.hpp"
+
+namespace gridtools {
+    namespace ghex {
+        namespace tl {
+            namespace mpi {
+                
+                /** @brief Class representing node (shared memory) topology. */
+                class hwtopo {
```
mbianco: If we'll use hwloc this name could become confusing. Should we switch to another name line "rank_topology" ?

include/ghex/transport_layer/ucx/context.hpp
```diff
@@ -87,6 +87,7 @@ namespace gridtools {
             private: // members
                 
                 thread_primitives_type&    m_thread_primitives;
+                mpi::hwtopo&               m_hwtopo;
```
mbianco: Is `hwtopo` modifiable? If not maybe this can be turned into
```suggestion
                mpi::hwtopo const&               m_hwtopo;
```

include/ghex/transport_layer/ucx/worker.hpp
```diff
@@ -98,6 +98,8 @@ namespace gridtools {
                     inline ucp_worker_h get() const noexcept { return m_worker.get(); }
                     address_t address() const noexcept { return m_address; }
                     inline const endpoint_t& connect(rank_type rank);
+
+                    const ::gridtools::ghex::tl::mpi::hwtopo& hwtopo() const noexcept { return m_context->m_hwtopo; }
```
mbianco: maybe a better name for this at this level? This is related to the naming question above

tests/transport/test_locality.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto& context = *context_ptr;
+    auto comm = context.get_communicator(context.get_token());
+
+    // test self
+    EXPECT_TRUE( comm.is_local(comm.rank()) );
+
+    // check for symmetry
+    std::vector<int> local_ranks(comm.size());
+    std::vector<char> my_host_name(HOST_NAME_MAX+1,0);
+    std::vector<char> other_host_name(HOST_NAME_MAX+1,0);
```
mbianco:
```suggestion
    // host names must be contained in a message-compatible data
    std::vector<char> my_host_name(HOST_NAME_MAX+1,0);
    std::vector<char> other_host_name(HOST_NAME_MAX+1,0);
```
Just adding the comment, since I spent some time on why this was a vector and not just a `char[]`

boeschf: no reason, will change

include/ghex/transport_layer/ucx/worker.hpp
boeschf: ok

include/ghex/transport_layer/ucx/context.hpp
boeschf: sure

include/ghex/transport_layer/mpi/hwtopo.hpp
boeschf: yes, will do

PR #77, opened 2020-03-23
added docker file to compile and run tests of GHEX

mbianco:
This is my first dockerfile. I want to use this to run Travis CI to test basic GHEX settings

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,73 @@
+ARG BUILD_FROM=ubuntu:19.10
+FROM ${BUILD_FROM}
+LABEL maintainer="Mauro Bianco <bianco@cscs.ch>"
+
+RUN apt-get update -qq && \
+    apt-get install -qq -y \
+    build-essential \
+    wget \
+    git \
+    tar \
+    software-properties-common && \
+    rm -rf /var/lib/apt/lists/*
+
+ARG CMAKE_VERSION=3.14.5
+RUN wget -q https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz && \
+    tar xzf cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz && \
+    cp -r cmake-${CMAKE_VERSION}-Linux-x86_64/bin cmake-${CMAKE_VERSION}-Linux-x86_64/share /usr/local/ && \
+    rm -rf cmake-${CMAKE_VERSION}-Linux-x86_64*
+
+ARG BOOST_VERSION=1.67.0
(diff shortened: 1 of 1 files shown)


mbianco:
@fthaler Just in case you want to comment on this file...

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+     
+## RUN wget -q http://www.mpich.org/static/downloads/3.1.4/mpich-3.1.4.tar.gz && \
+##     tar xf mpich-3.1.4.tar.gz && \
+##     cd mpich-3.1.4 && \
+##     ./configure --with-device=ch3:sock --disable-fortran --enable-fast=all,O3 --prefix=/usr && \
+##     make -j $(nproc) && \
+##     make install && \
+##     ldconfig && \
+##     cd .. && \
+##     rm -rf mpich-3.1.4*
```
fthaler: There would be one advantage of using this MPICH version: you could run your tests in the _same container_ on Piz Daint using Daint’s native MPI. So you could quite easily set up CSCS-Jenkins using Sarus I guess.

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DGridTools_DIR=/usr/local/lib/cmake \
+    -DGHEX_BUILD_TESTS=ON \
+    -DMPIEXEC_PREFLAGS="--allow-run-as-root;--oversubscribe" \
+    .. && \
+    make -j $(nproc) install
+
+
+RUN cd /GHEX/build && \
+    ctest --verbose
```
fthaler: You seem to run tests twice: once while building the container and once when running it (below). Is this on purpose?

mbianco: I forgot to remove it from the building. I'm now trying a different approach, suggested by Hannes, to use github actions, and hence to build the container and then use it to run the tests. I'm taking inspiration from serial-box and gridtools-docker for this... still work to do , though

fthaler: You can run exactly the same stuff on Github actions as on Travis, so this doesn’t matter much. But the approach for GTBench was indeed chosen to provide container instances that can be pulled from docker hub and directly run on Piz Daint. For a library this is probably not required… And at some point I might also add a level of indirection to the GTBench containers to avoid re-building the whole container every time.

mbianco: My question is: what is the return code of the container execution and how is it returned? Is is as a normal command? I suppose so...

Dockerfile
mbianco: the problem with MPICH, on my VM, is that I could not run multiple processes on a single cpu. OpenMPI works instead, by using `--oversubscribe` option. Do you use MPICH in a oversubscribed way?

mbianco:
I'm closing this PR since we do it on githubactions now.

PR #78, opened 2020-03-26
Fix native MPI benchmarks

angainor:
add an option to use MPI_Test vs MPI_Testany and MPI_Wait vs MPI_Waitall
missing include file
utility function (in benchmarks) to allow thread to core binding - not used yet


benchmarks/transport/mpi_p2p_bi_avail_mt.cpp
```diff
@@ -13,10 +13,38 @@
...
 
-#include <ghex/transport_layer/ucx/threads.hpp>
 #include <ghex/common/timer.hpp>
 #include "utils.hpp"
 
+#ifdef USE_OPENMP
+#include <omp.h>
+#define DO_PRAGMA(x) _Pragma(#x)
+
+#define GET_THREAD_NUM()  omp_get_thread_num()
```
mbianco, line 23: I was trying to rewrite the benchmark using thread primitives. Not sure if it makes sense... What do you think guys?

benchmarks/transport/mpi_p2p_bi_avail_mt.cpp
```diff
@@ -156,5 +207,5 @@ int main(int argc, char *argv[])
     } 
 
     MPI_Barrier(MPI_COMM_WORLD);
-    MPI_Finalize();
+    // MPI_Finalize(); // segfault due to unfinished requests
```
mbianco, line 210: Don't we have any mean to clean this up for the benchmark?

benchmarks/transport/mpi_p2p_bi_avail_mt.cpp
angainor: @mbianco I wanted to keep the MPI benchmarks as stand-alone as possible, without any GHEX deps. there is the timer, but otherwise it should be clean now. These benchmarks are very simple, and I only need threads/no threads. Their usage is also quite limited (mostly to test the MPI with threading). So I would not spend too much time on them.

benchmarks/transport/mpi_p2p_bi_avail_mt.cpp
angainor: @mbianco That is a problem. We cannot cancel send requests, so in our ghex benchmarks I deal with this by a tail-loop, which waits for all send requests to complete, and they by canceling all recv requests. We could probably do this here as well, but it is a lot of code - actually more than the test loop itself. It was much simpler to simply not call `MPI_Finalize()` - so far I don't see any side-effects, except for maybe some warnings, which can be turned off.

But if you think it's worth it, we can probably make a 'correct' test.

mbianco: It is not clear to me what value a benchmark has if it does not represents code that can be written safely by a user...

angainor: The benchmark loop is important and representative of user application performance. The tail loops and cleanup loops are not part of the benchmark. Their only purpose is to exit the program.

In actual user application the users either send/recv _exactly_ the needed number of messages, all of which are always matched/completed (halo exchange in a static parallel environment) - and hence never end up in this problem, or they have to implement something similar to the tail-loops in order to handle rank migration etc (DISPATCH). If you pre-post a recv message, which will never arrive, you have to do something with it. This use-case is not frequent, application-specific, and we do not measure the time of that.

PR #80, opened 2020-04-01
Cubed sphere

boeschf:


include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+#endif
+
+                /** @brief Describes a field spanning a subdomain of a cubed-sphere tile
+                  * @tparam T value type
+                  * @tparam Arch Architecture type
+                  * @tparam ComponentOrder Data layout order parameter (3 == stride 1, 0 == largest stride)
+                  * @tparam XOrder Data layout order parameter (3 == stride 1, 0 == largest stride)
+                  * @tparam YOrder Data layout order parameter (3 == stride 1, 0 == largest stride)
+                  * @tparam ZOrder Data layout order parameter (3 == stride 1, 0 == largest stride)*/
+                template<typename T, typename Arch, int ComponentOrder=0, int XOrder=3, int YOrder=2, int ZOrder=1>
```
mbianco, line 135: why not using `layout_map`? 
This my default is for a 4D storage?

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+                class field_descriptor {
+                public: // member types
+                    using value_type             = T;
+                    using arch_type              = Arch;
+                    using device_id_type         = typename arch_traits<arch_type>::device_id_type;
+                    using domain_descriptor_type = domain_descriptor;
+                    using dimension              = domain_descriptor_type::dimension;
+                    using layout_map             = ::gridtools::layout_map<ComponentOrder,XOrder,YOrder,ZOrder>;
```
mbianco, line 143: Here we assume component order if the first component of the array accessor. Is this a specific FV3 convention?

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+                    using domain_id_type         = domain_id_t;
+                    using coordinate_type        = ::gridtools::array<
+                        typename domain_descriptor_type::coordinate_type::value_type, dimension::value>;
+                    using size_type              = unsigned int;
+                    using strides_type           = ::gridtools::array<size_type, dimension::value>;
+                    using serialization_type     = serialization<arch_type,layout_map>;
+
+                    // holds buffer information (read-write access)
+                    struct buffer_descriptor {
```
mbianco, line 152: Aren't those already defined somewhere similarly?

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+                    value_type*     m_data;          /// pointer to data
+                    coordinate_type m_dom_first;     /// global coordinate of first domain cell
+                    coordinate_type m_offsets;       /// offset from beginning of memory to the first domain cell
+                    coordinate_type m_extents;       /// extent of memory (including halos)
+                    device_id_type  m_device_id;     /// device id
+                    strides_type    m_byte_strides;  /// memory strides in bytes
+
+                public: // ctors
+                    /** @brief construct field_descriptor from existing data
+                      * @tparam Array 3-dimensional array-like type
```
mbianco, line 312: Is it 3D only?

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+                      * @param dom domain descriptor
+                      * @param data pointer to data
+                      * @param offsets (x,y,z)-offset of first phyisical value (not halo) from the orign of the memory
+                      * @param extents (x,y,z)-extent of the memory (including buffer regions)
```
mbianco, line 316:
```suggestion
                      * @param extents (x,y,z)-extent of the memory (including buffer/padding regions)
```

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+                      * @param n_components number of components in this field
+                      * @param is_vector whether this field is a vector or a collection of scalars
+                      * @param d_id device id */
+                    template<typename Array>
+                    field_descriptor(const domain_descriptor& dom, value_type* data,
+                                     const Array& offsets, const Array& extents, unsigned int n_components,
```
mbianco, line 322: This field type uses components, too. This field type, though, should be identical to the structured grid fields, right? Should we just keep one field descriptor for both cases? Or are the packing/unpacking functions specific to inflated-cube? Could we reuse at least part of the code?

include/ghex/structured/cubed_sphere/domain_descriptor.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include <cassert>
+#include "../../common/coordinate.hpp"
+
+namespace gridtools {
+    namespace ghex {
+        namespace structured {
+            namespace cubed_sphere {
+
+                /** @brief cube dimensions */
+                struct cube {
```
mbianco, line 24: What is this struct describing more precisely?

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+
+                /** @brief Describes a field spanning a subdomain of a cubed-sphere tile
+                  * @tparam T value type
+                  * @tparam Arch Architecture type
+                  * @tparam ComponentOrder Data layout order parameter (3 == stride 1, 0 == largest stride)
+                  * @tparam XOrder Data layout order parameter (3 == stride 1, 0 == largest stride)
+                  * @tparam YOrder Data layout order parameter (3 == stride 1, 0 == largest stride)
+                  * @tparam ZOrder Data layout order parameter (3 == stride 1, 0 == largest stride)*/
+                template<typename T, typename Arch, int ComponentOrder=0, int XOrder=3, int YOrder=2, int ZOrder=1>
+                class field_descriptor {
```
mbianco, line 136: Is this class equivalent to `simple_field_wrapper`? Wouldn't it make sense to have only one wrapper? Can this wrapper be used also for regular grids?

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+                    strides_type    m_byte_strides;  /// memory strides in bytes
+
+                public: // ctors
+                    /** @brief construct field_descriptor from existing data
+                      * @tparam Array 3-dimensional array-like type
+                      * @param dom domain descriptor
+                      * @param data pointer to data
+                      * @param offsets (x,y,z)-offset of first phyisical value (not halo) from the orign of the memory
+                      * @param extents (x,y,z)-extent of the memory (including buffer regions)
+                      * @param n_components number of components in this field
```
mbianco, line 317:
```suggestion
                      * @param n_components number of components in this vector field
```

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,504 @@
...
+                    bool            m_is_vector;     /// is this data describing a vector field 
+                    value_type*     m_data;          /// pointer to data
+                    coordinate_type m_dom_first;     /// global coordinate of first domain cell
+                    coordinate_type m_offsets;       /// offset from beginning of memory to the first domain cell
+                    coordinate_type m_extents;       /// extent of memory (including halos)
+                    device_id_type  m_device_id;     /// device id
+                    strides_type    m_byte_strides;  /// memory strides in bytes
+
+                public: // ctors
+                    /** @brief construct field_descriptor from existing data
```
mbianco, line 311:
```suggestion
                    /** @brief construct field_descriptor from existing data representing a vector-field
```

include/ghex/structured/simple_field_wrapper.hpp
```diff
@@ -296,6 +296,9 @@ namespace structured {
 
         GT_FUNCTION
         void set_data(value_type* ptr) { m_data = ptr; }
+        
+        GT_FUNCTION
+        int num_components() const noexcept { return 1; }
```
mbianco, line 301: This goes with other comments: would it make sense to have a single wrapper to handle vector fields?
Also, should we choose between `wrapper` of `descriptor` consistently?

include/ghex/structured/cubed_sphere/field_descriptor.hpp
boeschf: changed this to variadic template parameter pack - 3D and 4D possible

include/ghex/structured/cubed_sphere/field_descriptor.hpp
boeschf: no this is not a specific FV3 convention - I chose it (also it is now the last in the list). However, it is used internally only anyways: the user would rarely use our field_descriptor directly.

include/ghex/structured/cubed_sphere/field_descriptor.hpp
boeschf: I hope my refactoring has led to some more code reuse

include/ghex/structured/cubed_sphere/field_descriptor.hpp
boeschf: the domain id 3D always, the data can be 4D

include/ghex/structured/cubed_sphere/field_descriptor.hpp
boeschf: I gave it a try - it changed a lot of the code - but I believe it became more maintainable

include/ghex/structured/cubed_sphere/field_descriptor.hpp
boeschf: see earlier comments - I tried to consolidate and reuse parts of the code and packing/serialization

boeschf:
I am finally ready with the requested changes - I will add some more comments tonight.
things that changed:
- folder/namespace structure: split `structured` into `regular` and `cubed_sphere`
- code reuse
   - use base class for all structured field descriptors
   - use same packing kernels for all structured grids
- made interface of regular and cubed_sphere fields more symmetric
- iterator based interface to the communication object
- todo:
  - reintroduce specialization for gpu-packing in the special case of uniform field range (i.e. when only fields with same type are passed to the exchange, the gpu can be loaded with a few bigger kernels and thus we reduce the number of kernel launches)

include/ghex/structured/regular/field_descriptor.hpp
```diff
@@ -0,0 +1,177 @@
...
+    using size_type                = typename base::size_type;
+    using strides_type             = typename base::strides_type;
+    using serialization_type       = gridtools::ghex::structured::serialization<arch_type,layout_map>;
+    using pack_iteration_space     = typename base::pack_iteration_space;
+    using unpack_iteration_space   = typename base::unpack_iteration_space;
+
+public: // ctors
+    template<typename ExtentArray, typename OffsetArray>
+    field_descriptor(
+            const domain_descriptor_type& dom_,
```
mbianco, line 54: data members should use `m_`

include/ghex/structured/cubed_sphere/halo_generator.hpp
```diff
@@ -0,0 +1,281 @@
...
+                      * @param d domain descriptor: determines the coordinate system in which to perform the intersection
+                      * @param first_a_local first coordinate of the iteration space a in local coordinates
+                      * @param last_a_local last coordinate of the iteration space a in local coordinates
+                      * @param first_a_global first coordinate of the iteration space a in global coordinates
+                      * @param last_a_global last coordinate of the iteration space in global coordinates
+                      * @param first_b_global first coordinate of the iteration space b in global coordinates
+                      * @param last_b_global last coordinate of the iteration space b in global coordinates
+                      * @return  valid intersection (as pairs of local and global iteration spaces) of a and b if
+                      * b is on the same cube tile as the domain, invalid intersection otherwise */
+                    box2 intersect(const domain_type& d,
```
bettiolm, line 218: Where is this used?

include/ghex/communication_object_2.hpp
```diff
@@ -267,89 +266,92 @@ namespace gridtools {
...
+//                using field_type = std::remove_reference_t<decltype(first->get_field())>;
+//                using value_type = typename field_type::value_type;
+//                auto h = exchange_impl(first, length);
+//                post_recvs();
+//                h.m_wait_fct = [this](){this->wait_u<value_type,field_type>();};
+//                memory_t& mem = std::get<memory_t>(m_mem);
+//                packer<gpu>::template pack_u<value_type,field_type>(mem, m_send_futures, m_comm);
+//                return h;
+//            }
+//#endif
```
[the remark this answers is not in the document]
boeschf: I'll leave those as a reminder to re-enable the optimized communication for same-type fields on gpus

include/ghex/structured/cubed_sphere/domain_descriptor.hpp
```diff
@@ -0,0 +1,110 @@
...
+                
+                static inline bool operator==(const domain_id_t& a, const domain_id_t& b) noexcept {
+                    return a.tile==b.tile && a.id==b.id;
+                }
+
+                /** @brief implements domain descriptor concept for cubed sphere grids */
+                class domain_descriptor
+                {
+                public: // member types
+                    using domain_id_type      = domain_id_t;
```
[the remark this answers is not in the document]
boeschf: renamed

include/ghex/structured/cubed_sphere/domain_descriptor.hpp
```diff
@@ -0,0 +1,110 @@
...
+                        m_first[1] = x_first;
+                        m_last[1]  = x_last;
+                        m_first[2] = y_first;
+                        m_last[2]  = y_last;
+                        m_first[3] = 0;
+                        m_last[3]  = m_c.levels-1;
+                    }
+
+                public: // member functions
+                    int x() const noexcept { return m_c.x; }
```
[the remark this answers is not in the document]
boeschf: renamed

include/ghex/structured/cubed_sphere/field_descriptor.hpp
```diff
@@ -0,0 +1,226 @@
...
+            else if (m_is_vector && coord[3]==1 && m_transform.m_rotation[1]==-1)
+                return -(*reinterpret_cast<const T*>(reinterpret_cast<const char*>(
+                    base_is::m_buffer_desc.m_ptr) + memory_location));
+            else
+                return *reinterpret_cast<const T*>(reinterpret_cast<const char*>(
+                    base_is::m_buffer_desc.m_ptr) + memory_location);
+        }
+    };
+
+    int             m_c;             /// cube size
```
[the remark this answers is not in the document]
boeschf: renamed

include/ghex/structured/field_utils.hpp
```diff
@@ -13,8 +13,55 @@
 
 #include <gridtools/common/host_device.hpp>
 #include <cstddef> 
+#include <gridtools/common/array.hpp>
 
 namespace gridtools {
+
+    template <typename T, size_t D>
+    GT_FUNCTION
+    array<T,D> operator+(array<T,D> a, const array<T,D>& b)
```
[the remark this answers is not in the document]
boeschf: I make a copy of the first operand and modify it. Maybe we need to add our own fixed-size vector/coordinate class at some point to get rid of gridtools dependency here and avoid any possible dangers.

include/ghex/structured/cubed_sphere/domain_descriptor.hpp
```diff
@@ -0,0 +1,110 @@
...
+                struct domain_id_t {
+                    int tile;
+                    coordinate<std::array<int,2>> id;
+                };
+
+                // strict ordering of domain_id objects
+                static inline bool operator>(const domain_id_t& a, const domain_id_t& b) noexcept {
+                    if (a.tile > b.tile) return true;
+                    if (b.tile > a.tile) return false;
+                    if (a.id[1] > b.id[1]) return true;
```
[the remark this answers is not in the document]
boeschf: yes, it is arbitrary. but it must be a strict (global) ordering.

include/ghex/structured/cubed_sphere/domain_descriptor.hpp
```diff
@@ -0,0 +1,110 @@
...
+                }
+
+                /** @brief implements domain descriptor concept for cubed sphere grids */
+                class domain_descriptor
+                {
+                public: // member types
+                    using domain_id_type      = domain_id_t;
+                    // 4-dimensional: tile, x, y, z
+                    using dimension           = std::integral_constant<int,3>;
+                    using coordinate_type     = coordinate<std::array<int,4>>;
```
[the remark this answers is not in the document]
boeschf: yes, you are right. I could possibly pull out the z coordinate and store/treat it separately

include/ghex/structured/cubed_sphere/halo_generator.hpp
```diff
@@ -0,0 +1,281 @@
...
+                    std::array<int, 4> m_halo;
+
+                public: // ctors
+                    /** @brief add an equal halo around the domain (default case)
+                      * @param halo number of halo cells */
+                    halo_generator(int halo) noexcept : m_halo{halo,halo,halo,halo} {}
+
+                    /** @brief construct with custom halo
+                      * @tparam Array 4-D array type
+                      * @param halos halos in +x,-x,-y,+y directions (for staggered grids) */
```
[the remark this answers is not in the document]
boeschf: for staggered grids you might want to use different halos on each side of the grid.
In general, for unstructured grids (which this kind of is when projected to the sphere) it makes sense to just say how many neighbour cells we should consider (i.e. 1 number). In the special case of cubed-sphere grids, it can be useful to specify halos per direction to easily enable staggering.

include/ghex/structured/cubed_sphere/halo_generator.hpp
```diff
@@ -0,0 +1,281 @@
...
+                    auto operator()(const domain_type& d) const {
+                        // get tile id
+                        const int tile_id = d.domain_id().tile;
+                        // get resolution of cube
+                        const int c = d.x();
+                        // domain in global tile coords
+                        const box domain_box{ d.first(), d.last() };
+                        // domain in local tile coords
+                        const box domain_box_local{
+                            coordinate_type{0/*tile_id*/, 0, 0, 0},
```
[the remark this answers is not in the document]
boeschf: yes

include/ghex/structured/cubed_sphere/halo_generator.hpp
```diff
@@ -0,0 +1,281 @@
...
+                            coordinate_type{0/*tile_id*/, d.last()[1]-d.first()[1],
+                                                     d.last()[2]-d.first()[2], 
+                                                     d.last()[3]-d.first()[3]}}; 
+                        // cube tile in global tile coords
+                        const box tile_box{coordinate_type{tile_id,  0,  0,d.first()[3]},
+                                           coordinate_type{tile_id,c-1,c-1,d.last()[3]}};
+                        // create 4 neighbor regions based on neighbor tiles
+                        // in global tile coords
```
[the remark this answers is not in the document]
boeschf: the `tile_box` above is my current tile described in global coordinates. From this temporary variable I create the receiving halo regions in "fake" global coordinates (fake in the sense that they don't exist as a real tile). So, yes you are completely correct.

include/ghex/structured/regular/domain_descriptor.hpp
```diff
@@ -0,0 +1,72 @@
...
+        using dimension           = std::integral_constant<int,Dimension>;
+        using coordinate_type     = std::array<int,dimension::value>;
+
+    public: // ctors
+        /** @brief construct a local domain
+         * @tparam Array coordinate-like type
+         * @param id domain id
+         * @param first first coordinate in domain (global coordinate)
+         * @param last last coordinate in domain (including, global coordinate) */
+        template<typename Array>
```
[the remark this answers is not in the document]
boeschf: maybe we can remove it, but some applications may for example want to use a coordinate as domain id and some others scalars. But maybe in a further round of simplification we can remove this template.

include/ghex/structured/field_descriptor.hpp
```diff
@@ -0,0 +1,254 @@
...
+template<typename T, typename Arch, typename DomainDescriptor, int... Order>
+class field_descriptor
+{
+public: // member types
+    using value_type             = T;
+    using arch_type              = Arch;
+    using device_id_type         = typename arch_traits<arch_type>::device_id_type;
+    using domain_descriptor_type = DomainDescriptor;
+    using dimension              = std::integral_constant<std::size_t,sizeof...(Order)>;
+    using has_components         = std::integral_constant<bool,
```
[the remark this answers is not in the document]
boeschf: - in case the field is 3 dimensional (usual case), we have for example `field_descriptor<...,2,1,0>`, so 3 order arguments, this leads to dimension being `3`
    - the cubed sphere domain has a dimension of `3` (see `cubed_sphere::domain_descriptor::dimension`)
    - then `has_components = 3 > 3 = false`
- when we have a field with components we would use `field_descriptor<...,2,1,0,3>` for example
    - then `has_components = 4 > 3 = true`

include/ghex/structured/field_descriptor.hpp
```diff
@@ -0,0 +1,254 @@
...
+            const coordinate_type data_coord = coord + m_data_is.m_offset;
+            // dot product with strides to compute address
+            const auto memory_location = dot(m_data_is.m_strides, data_coord);
+            return *reinterpret_cast<T*>(
+                reinterpret_cast<char*>(m_data_is.m_ptr) + memory_location);
+        }
+    };
+
+protected: // members
+    domain_descriptor_type m_dom;             ///< domain descriptor
```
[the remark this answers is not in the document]
boeschf: good question. I don't see a problem with a copy here (it's probably even safer from a dangling reference point) since the domain descriptor does not allocate and is cheap to copy.

include/ghex/structured/cubed_sphere/halo_generator.hpp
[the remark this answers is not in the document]
bettiolm: Right, I see

include/ghex/structured/field_descriptor.hpp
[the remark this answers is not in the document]
bettiolm: Ok, sure.

PR #81, opened 2020-04-03
reduced the length of the tests, since containers seems to be pretty …

mbianco:
…slow


mbianco:
I merged this in my first workflow file. I think we can make the tests longer again, but not sure...

PR #84, opened 2020-04-10
Question: changes to enable spinlock on macos

mbianco:
Do we want this in the repo? I had to put this code from somewhere to get the code compiling and running on my Mac.

### DIFF
--- a/include/ghex/threads/mutex/pthread_spin/mutex.hpp
+++ b/include/ghex/threads/mutex/pthread_spin/mutex.hpp
@@ -1,12 +1,12 @@
-/* 
+/*
  * GridTools
- * 
+ *
  * Copyright (c) 2014-2020, ETH Zurich
  * All rights reserved.
- * 
+ *
  * Please, refer to the LICENSE file in the root directory.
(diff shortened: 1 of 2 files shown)


angainor:
Wow. Are there no pthread spin-locks on macos? I guess we can have it, but then again I guess in a case where something is not supported on a given platform we can simply use a different set of thread-primitives? mutexes instead of spin locks?

mbianco:
MacOS is not a target platform, unless Cray decides to be cool 😎 . So I would put this hack just for development

include/ghex/threads/mutex/pthread_spin/mutex.hpp
```diff
@@ -15,6 +15,45 @@
 #include <pthread.h>
 #include <sched.h>
 
+#ifdef _MACOS_GHEX_
+typedef int pthread_spinlock_t;
```
mbianco, line 19:
```suggestion
// This code was cut&pasted from stackoverflow - copyright of the person posting it
typedef int pthread_spinlock_t;
```

PR #85, opened 2020-04-15
CTest dashboard integration and config

pycicle:


biddisco:
If possible, please merge this small PR quickly, once this is in, then I can enable CDash dashboard testing for the main GridTools/GHEX project at https://cdash.cscs.ch/index.php?project=GHEX - which is currently testing only PRs on my fork of the repo.

biddisco:
The existing `enable_testing` sets up things(commands) like `add_test` and suchlike for testing. The only real addition here is the `include(CTest)` which also loads support for `CDash` submissions using the settings in the config file supplied - the catch is that `include(CTest)` actually calls `enable_testing` which adds the `BUILD_TESTING` cmake flag, so really you could remove the `enable_testing` - but you couldn't use
```
if(BUILD_TESTING)
  include(CTest) # this internally calls enable_testing()
endif()
```
because the `BUILD_TESTING` doesn't exist until the include is included.
```
enable_testing()
if(BUILD_TESTING)
  include(CTest) # this internally calls enable_testing()
endif()
```
works, but the overhead of CTest is not worth the extra `if` block to be honest and everyone just uses
```
enable_testing()
include(CTest)
```
even though 
```
include(CTest)
```
is enough.

CMakeLists.txt
```diff
@@ -94,6 +94,10 @@ if (GHEX_ENABLE_ATLAS_BINDINGS)
 endif()
 target_compile_features(ghexlib INTERFACE cxx_std_14)
 
+# Enable CDash dashboard testing integration
+include(CTest)
+
+# Enable adding of tests etc
 enable_testing()
```
mbianco, line 101:
```suggestion
# Enable adding of tests etc
enable_testing()

# Enable CDash dashboard testing integration
include(CTest)
```

Should it be like this then?

mbianco: @biddisco any feedback?

biddisco: Doesn't really make any difference which way around you put them. Personally I prefer the way you've done it, but tons of other projects do it the other way around (I'm guessing because cut'n'paste is quite common in cmake scripts and nobody really cares).

mbianco: I'll merge the suggestion then and merge

PR #88, opened 2020-04-29
Unstructured

bettiolm:
Major changes:
- provided unstructured user concepts with a common interface for handling general unstructured grids. Domains can be constructed from a mapping between grid vertices and their adjacency;
- support for domain over-subscription (multiple domains per rank, different number of domains per rank);
- ids of neighbour domains which to receive from, together with their rank location, can optionally be provided through a user-defined `recv_domain_ids_generator`, in order to optimize the collective communications in the setup phase;
- atlas user concepts, tests and benchmark updated accordingly.


angainor:
@bettiolm @mbianco I have some general questions about the unstructured grids to make sure I understand it. I see in the code mentions of ParMetis native storage (CSR), or in other (Petsc) terminology - CRS. But I also see that there is another format of storage, which GHEX uses internally - `map_type = std::vector<std::pair<global_index_type, adjncy_type>>;`. It also says in the code that 

```
adjncy_type m_adjncy; // named after ParMetis CSR arrays; TO DO: needed mostly for debugging, can be removed
```
So the question is: is it possible for the user to use the CRS storage? By that I mean to pass the CRS pointers to GHEX at init, but also for GHEX to use those pointers internally - if needed. The problem is that in my previous experience with this type of methods the sparse storage is the dominant memory component. If we then insist that there are two copies of the matrix stored in memory, then we effectively cut down the problem size that can be solved by half. On the other hand, I would imagine that GHEX doesn't need to store all domain connectivity data (full sparse matrix), but only the halo information. Then the memory overhead would not be large. But from the code it seems to me that we keep info about the entire domain. Could you briefly tell me how that is done?

Another thing is the Blocked CRS storage. In this case the adjacency matrix identifies dense non-zero sub-blocks in the matrix (e.g., for 3D mechanics we have 3 velocities vx, vy, vz, and the matrix blocks are 3x3). In this case - for performance reasons and for memory reasons - the Blocked storage is used. So instead of identifying each non-zer entry with a column index, you identify blocks of entries. The storage is the same from the connectivity perspective, but requires slightly different code for halo exchange itself: essentially re-scaling by the block size when you access the vectors. Do we (plan to) support this in GHEX?

bettiolm:
> @bettiolm @mbianco I have some general questions about the unstructured grids to make sure I understand it. I see in the code mentions of ParMetis native storage (CSR), or in other (Petsc) terminology - CRS. But I also see that there is another format of storage, which GHEX uses internally - `map_type = std::vector<std::pair<global_index_type, adjncy_type>>;`. It also says in the code that
> 
> ```
> adjncy_type m_adjncy; // named after ParMetis CSR arrays; TO DO: needed mostly for debugging, can be removed
> ```
> 
> So the question is: is it possible for the user to use the CRS storage? By that I mean to pass the CRS pointers to GHEX at init, but also for GHEX to use those pointers internally - if needed. The problem is that in my previous experience with this type of methods the sparse storage is the dominant memory component. If we then insist that there are two copies of the matrix stored in memory, then we effectively cut down the problem size that can be solved by half. On the other hand, I would imagine that GHEX doesn't need to store all domain connectivity data (full sparse matrix), but only the halo information. Then the memory overhead would not be large. But from the code it seems to me that we keep info about the entire domain. Could you briefly tell me how that is done?
> 
> Another thing is the Blocked CRS storage. In this case the adjacency matrix identifies dense non-zero sub-blocks in the matrix (e.g., for 3D mechanics we have 3 velocities vx, vy, vz, and the matrix blocks are 3x3). In this case - for performance reasons and for memory reasons - the Blocked storage is used. So instead of identifying each non-zer entry with a column index, you identify blocks of entries. The storage is the same from the connectivity perspective, but requires slightly different code for halo exchange itself: essentially re-scaling by the block size when you access the vectors. Do we (plan to) support this in GHEX?

@angainor on your first point: what is strictly needed in the domain descriptor, at least as it is thought right now, is a data structure which stores in a given order both the vertices that belong to the domain and the halo points. In my current implementation, for example, I have a single `std::vector` which stores the `n` inner vertices in term of global coordinates (indices [0, n-1]) and, contiguously, the halo vertices (indices [n, size-1]). This is deduced from the connectivity information, but the connectivity itself is not stored (yes, I should remove `m_adjncy` member actually, it is there just because I used it for debugging). The `map_type` data structure you refer to is not stored in the domain descriptor. It is just a facility used in one of the constructor overloads and can be constructed at the application level from c-like CSR arrays. That still involves a copy anyway, and I totally agree with you that it is not optimal in terms of memory requirements. Anyway I think this can be easily avoided just providing an overload which directly accepts references to the original arrays if we need. This should not be too complex.

Additional note: the halo information which is required by GHEX involves only receive halos, and therefore I have the domain descriptor storing all internal points as well, since in principle any of those may be accessed in a send operation. I don't think this could be easily changed, at least without changing the library requirements.

On the second point: maybe some optimization for blocked storage, at least in terms of performances, can be done at the access operators level: a default implementation of these operators is provided in the unstructured user concepts code included in this PR, but as usual, the user is free to use custom ones in case he is aware of specific optimizations. Anyway I agree that probably that wouldn't be enough. Providing a more complex implementation which optimizes the internal storage as well for the blocked case might be useful and we can definitely discuss it.

include/ghex/unstructured/grid.hpp
```diff
@@ -19,15 +19,15 @@ namespace gridtools {
...
                 template<typename Index>
                 struct grid {
-                    using index_t = Index;
+                    using index_type = Index;
                 };
 
             } // namespace detail
 
             /** @brief type to indicate unstructured grids */
             struct grid {
```
mbianco, line 28: It seems strange to have this structure here. Where is it used? Could this be simplified?

include/ghex/unstructured/pattern.hpp
```diff
@@ -191,129 +175,435 @@ namespace gridtools {
...
-                            }
+                        rank_local_start_index += halo_size;
+                    }
+
+                    return pattern_container<communicator_type, grid_type, domain_id_type>(std::move(my_patterns), m_max_tag);
+
+                }
+
+                template<typename Transport, typename ThreadPrimitives, typename HaloGenerator, typename RecvDomainIdsGen, typename DomainRange>
+                static auto apply(tl::context<Transport, ThreadPrimitives>& context, HaloGenerator&& hgen, RecvDomainIdsGen&& recv_domain_ids_gen, DomainRange&& d_range) {
```
mbianco, line 435: These functions are huge. Is it possible to write somewhere a description to understand what's going on in an intuitive way?

include/ghex/unstructured/grid.hpp
bettiolm: Here my attempt is to be consistent with the type definitions used in the other pattern specialization (structured) and in the `make_pattern` function. In general, I would say that this code structure has the advantage that the user needs to specify only the grid type (choosing between the `structured` or `unstructured` struct) when calling the `make_pattern` function at the application level. The index type is then inferred directly from the domain (which needs to provide the member type `local_index_type`), and this ensures that the index type used in the Iteration space is the same used by the domain. In the unstructured case, with index type I mean the local index type.

include/ghex/unstructured/pattern.hpp
bettiolm: Added some documentation for these two functions. Can go more in detail if needed.

include/ghex/unstructured/grid.hpp
mbianco: But why defer to a `detail_` namespace? Could this be done with just one trait class?

bettiolm: Yes, I think so, but actually I do not see a way to do this without using the domain type as a template in the grid structure, which translates into an additional template parameter for the user. It could also be that I am not seeing a simpler solution, though. Anyway, modifying it will change the code in the structured specialization as well, (the same structure is used there as well).

include/ghex/unstructured/grid.hpp
```diff
@@ -19,15 +19,15 @@ namespace gridtools {
...
                 template<typename Domain>
-                using type = detail::grid<typename Domain::index_t>;
+                using type = detail::grid<typename Domain::local_index_type>;
```
mbianco, line 30:
```suggestion
                using index_type = typename Domain::local_index_type;
```
and instead of doing `grid::type::index_type` you just write `grid::index_type`. Am I missing something?

bettiolm: Yep, my concern would only be on the typedef inside the high level `make_pattern` (`./include/ghex/pattern.hpp:122`). I can use `using grid_type = GridType`, but then I would need the domain type to infer `index_type`, and this is not known i.e. inside the specialized `pattern.hpp` class (see for ex `./include/ghex/unstructured/pattern.hpp:40`). I can use `index_type` to chose among the `make_pattern` specializations, but in this way I am no longer discriminating among structured / unstructured. I might be a bit biased, though :)

mbianco: we can either have the grid containing the domain and the domain the index
```
template <class Domain>
struct grid {
    using domain_type = Domain;
};
```
or we can have metafunctions to deduce the Domain:
```
template <typename T>
struct domain_of;

template <typename Domain>
struct domain_of< grid<Domain> > {
    using type = Domain;
};
```

mbianco: I guess the question is: id the index a property of the grid or of the domain?

bettiolm: I think it should be a property of the Domain. My idea is: in the unstructured case by index type we mean the local index type. In the structured case the equivalent is the coordinate type. I think that in both cases this should be set on the application side (an thus defined in the domain descriptor) and GHEX should provide iteration spaces using the chosen coordinate / index type coherently.

bettiolm: I think what I was not sure of was whether it would be fine to explicitly template the grid type on the domain type or not.

mbianco: Le my try to reason about it. A field is where we store data, and a field relates to the domain, I guess. A field has a storage, and this can be 1-2-3-4-5D whatever, even though the access function for the field entity should always be based on domain::index somehow. Either we say the partitioned grid defines the (partitioned) domain, or that the domain defines the grid, they seem pretty tight to one another to me. The grid has additional relational information that defines the neighborhood, while the domain is just the "space" where the fields live. The domain asks the grid for neighbors, I guess, so Maybe the domain should be derived from the Grid, and the fields from the domain...

include/ghex/unstructured/user_concepts.hpp
```diff
@@ -0,0 +1,335 @@
...
+                        assert(field.size() == (domain.size() * domain.levels()));
+                    }
+
+                    // member functions
+
+                    device_id_type device_id() const noexcept { return 0; } // significant for the GPU
+                    domain_id_type domain_id() const noexcept { return m_domain_id; }
+                    std::size_t domain_size() const noexcept { return m_domain_size; }
+                    std::size_t levels() const noexcept { return m_levels; }
+
```
bettiolm, line 270:
```suggestion
int num_components() const noexcept { return 1; }
```

bettiolm, line 270:
```suggestion
int num_components() const noexcept { return 1; }

```

include/ghex/unstructured/user_concepts.hpp
```diff
@@ -0,0 +1,336 @@
...
+                        assert(field.size() == (domain.size() * domain.levels()));
+                    }
+
+                    // member functions
+
+                    device_id_type device_id() const noexcept { return 0; } // significant for the GPU
+                    domain_id_type domain_id() const noexcept { return m_domain_id; }
+                    std::size_t domain_size() const noexcept { return m_domain_size; }
+                    std::size_t levels() const noexcept { return m_levels; }
+int num_components() const noexcept { return 1; }
```
bettiolm, line 270:
```suggestion
                    int num_components() const noexcept { return 1; }
```

PR #97, opened 2020-07-07
GridTools 2.0 Support

fthaler:
GT2.0 support, required at some point (soon) for the move of GTBench from GT1.1 to GT2. This drops support for GT1.1.


include/ghex/glue/gridtools/field.hpp
```diff
@@ -63,60 +63,53 @@ namespace gridtools {
...
-            struct get_arch
-            {
-                using type = ::gridtools::ghex::cpu;
-            };
 #ifdef __CUDACC__
+            // TODO: get proper arch from data store?
             template<typename DataType>
-            struct get_arch<cuda_storage<DataType>>
+            struct get_arch
             {
```
boeschf, line 70: I don't think hard coding the architecture to gpu when compiling with nvcc is a good idea w.r.t. hybrid usage (even though GridTools may not (yet?) support that).

utils/mpi_listener.hpp
```diff
@@ -0,0 +1,161 @@
+/*
```
mbianco, line 1: Didn't we already have this (and more than one version, too)?

include/ghex/glue/gridtools/field.hpp
mbianco: That makes sense. Why was this changed?

fthaler: Look at the TODO there comment there, I am fully aware of this ;) It was changed because the old code does not work anymore due to the changes in the GT API.

utils/mpi_listener.hpp
fthaler: Nope, was previously used from GT, but is not provided in the GT installation/public API anymore.

include/ghex/glue/gridtools/field.hpp
mbianco: Can we get the right solution here? Maybe we should ask Anton?

fthaler: I am sure there is some solution. If Anton has some time, it would be great if he could look into it as I am quite busy with other stuff right now :+1: @anstaf

anstaf: Firstly, this will not be compiled with `nvcc` and `GT2` --  there is no template named `cuda_storage` in `GT2`.
The second thing -- it is not a good idea to couple a GHEX entity with a storage type from GT: GT2 storage API is extensible; the traits that model a concept are passed to the builder. Even though there is predefined `gridtools::storage::gpu` traits the user can define its own. 
For generality I would suggest to pass `GHEX` Arch explicitly to the  `wrap_gt_field`.

fthaler: Thanks for looking into this! Manually passing the GHEX arch would be fine with me, I can add that if everyone agrees.

PR #100, opened 2020-08-14
In-place receive for unstructured

bettiolm:


include/ghex/communication_object_ipr.hpp
```diff
@@ -0,0 +1,564 @@
...
+ * All rights reserved.
+ * 
+ * Please, refer to the LICENSE file in the root directory.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * 
+ */
+#ifndef INCLUDED_GHEX_COMMUNICATION_OBJECT_IPR_HPP
+#define INCLUDED_GHEX_COMMUNICATION_OBJECT_IPR_HPP
+
+// TO DO: remove unneeded includes
```
mbianco, line 14: Check this before merging

include/ghex/communication_object_ipr.hpp
```diff
@@ -0,0 +1,564 @@
...
+#include <unordered_map>
+#include <queue>
+#include <stdio.h>
+#include <functional>
+
+namespace gridtools {
+
+    namespace ghex {
+
+        // namespace unstructured {
```
mbianco, line 32: remove? Ho do we protect against mis-use?

include/ghex/communication_object_ipr.hpp
```diff
@@ -0,0 +1,564 @@
...
+                    // member types
+
+                    using address_type = typename communicator_type::address_type;
+                    using index_container_type = typename pattern_type::index_container_type;
+                    using pack_function_type = std::function<void(void*, const index_container_type&, void*)>; // last argument intended as CUDA stream
+
+                    /** @brief pair of domain ids + tag id with ordering */
+                    struct domain_id_pair_and_tag {
+
+                            domain_id_type first_id;
```
mbianco, line 130: identation

tests/unstructured_user_concepts.cpp
```diff
@@ -486,13 +522,87 @@ TEST(unstructured_user_concepts, data_descriptor) {
     EXPECT_NO_THROW(co.bexchange(patterns(data)));
 
     auto h = co.exchange(patterns(data));
+    //auto h = co.exchange_ipr(patterns(data));
```
mbianco, line 525:
```suggestion
```

mbianco (commented):
I'm very slow on this one. Some  minor comments for now

include/ghex/communication_object_ipr.hpp
```diff
@@ -0,0 +1,564 @@
...
+                                    (first_id > other.first_id ? false :
+                                        (second_id < other.second_id ? true :
+                                            (second_id > other.second_id ? false :
+                                                (tag < other.tag)))));
+                            }
+
+                    };
+
+                    /** @brief Holds a pointer to a set of iteration spaces and a callback function pointer
+                     * which is used to store a field's pack or unpack member function.
```
mbianco, line 145: are these ones used for unpack? I guess not.

include/ghex/communication_object_ipr.hpp
```diff
@@ -0,0 +1,564 @@
...
+                     * This class also stores the offset in the serialized buffer in bytes.
+                     * The type-erased field_ptr member is only used for the gpu-vector-interface. // TO DO: verify is still needed in this case
+                     * @tparam Function either pack or unpack function pointer type */
+                    template<typename Function>
+                    struct field_info {
```
mbianco, line 150: Also this struct is the same as the co_2. Would it make sense to use some composition to reuse some code? (how much code would be saved by doing so?)

include/ghex/communication_object_ipr.hpp
```diff
@@ -0,0 +1,564 @@
...
+                            Function call_back;
+                            const index_container_type* index_container;
+                            std::size_t offset;
+                            void* field_ptr;
+                    };
+
+                    /** @brief Holds serial buffer memory and meta information associated with it.
+                     * @tparam Vector contiguous buffer memory type
+                     * @tparam Function pack function pointer type (not used for unpacking)*/
+                    template<class Vector, class Function>
```
mbianco, line 161: This struct also is the same as co_2. Maybe instead of composing classes, we can put these sub-structs into an impl namespace in a separate file and use them when we need them

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -0,0 +1,573 @@
...
+
+                private:
+
+                    // member types
+                    using communicator_type = Communicator;
+                    using index_type = Index;
+                    using grid_type = unstructured::detail::grid<index_type>;
+                    using domain_id_type = DomainIdType;
+
+                    // friend class // TO DO: is friend needed?
```
boeschf, line 53: the friend is needed because of private ctor

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -0,0 +1,573 @@
...
+                    friend class communication_object_ipr<communicator_type,grid_type,domain_id_type>;
+
+                    // members
+                    communicator_type m_comm; // TO DO: is it needed?
```
boeschf, line 57: comm is probably not needed

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -0,0 +1,573 @@
...
+                            for (auto& p0 : m.recv_memory) {
+                                for (auto& p1 : p0.second) {
+                                    p1.second.size = 0;
+                                    p1.second.field_infos.resize(0);
+                                }
+                            }
+                        });
+                    }
+
+                    // TO DO: double check from here on. First of all, some code could be avoided
```
boeschf, line 382: is this resolved?

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Right. Done.

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Removed from class members.

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Yes, I just left it behind

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -0,0 +1,561 @@
...
+                     * @tparam Function pack function pointer type (not used for unpacking)*/
+                    template<class Vector, class Function>
+                    struct buffer {
+                            using field_info_type = field_info<Function>;
+                            address_type address;
+                            int tag;
+                            Vector buffer;
+                            std::size_t size;
+                            std::vector<field_info_type> field_infos;
+                            cuda::stream m_cuda_stream;
```
bettiolm, line 160: Is it ok to stick with this naming convention (same used in `communication_object_2`)?

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -0,0 +1,561 @@
...
+                     * Vector = ipr_message, Function = std::function<void()>;
+                     * not a specialization, because it is no longer a buffer*/
+                    struct recv_ipr_info {
+                            using field_info_type = field_info<std::function<void()>>;
+                            address_type address;
+                            int tag;
+                            ipr_message message;
+                            std::size_t size;
+                            std::vector<field_info_type> field_infos;
+                            cuda::stream m_cuda_stream;
```
bettiolm, line 198: Same comment here on the naming convention

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -0,0 +1,561 @@
...
+
+                    void post_recvs() {
+                        detail::for_each(m_mem, [this](auto& m) {
+                            for (auto& p0 : m.recv_memory) {
+                                for (auto& p1: p0.second) {
+                                    if (p1.second.size > 0u) {
+                                        p1.second.message.resize(p1.second.size);
+                                        m.m_recv_futures.emplace_back(
+                                            typename std::remove_reference_t<decltype(m)>::future_type{
+                                                m_comm.recv(p1.second.message, p1.second.address, p1.second.tag).m_handle
```
bettiolm, line 322: Here I used the same constructor as in `communication_object_2`, i.e. the one with `handle` as argument, which moves only the handle. Not a big deal, but I think that using directly the future move constructor could also be an option, since we are using the same future specialization in both cases anyway. Probably this could be a bit too strict if one day we should decide (for some reason) not to use the `future<void>` specialization as return type for the send / recv functions.

include/ghex/unstructured/communication_object_ipr.hpp
mbianco: I did not notice the names here. We use `m_` for members that are not assumed to be accessed publicly.
Anyway, since these members are not accessed in this PR, I would leave the code as is, since otherwise you have to make more diffs.

include/ghex/unstructured/communication_object_ipr.hpp
mbianco: as above

include/ghex/unstructured/communication_object_ipr.hpp
mbianco: I'm not sure I understand completely. The trade-off between reuse of patterns and code is tricky. I would not spend too much thinking about possible future choices, and maybe put a comment explaining what to do if an error appears about it

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Ok, sure

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Yep

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Ok, I just added a comment in the code

include/ghex/communication_object_ipr.hpp
bettiolm: Done, added a couple of issues as well.

include/ghex/communication_object_ipr.hpp
bettiolm: Done, used a specialization for unstructured.

include/ghex/communication_object_ipr.hpp
bettiolm: Done

tests/unstructured_user_concepts.cpp
bettiolm: Done

include/ghex/communication_object_ipr.hpp
bettiolm: This struct is actually used for the receive part as well, using `std::function<void()>` as function type. I first thought of not using a func at all, but this requires at least to define a new struct (or a specialization of this one), and I think that keeping it as it is is less confusing in the end. I'll mention this in one of the two opened issues anyway.

include/ghex/communication_object_ipr.hpp
bettiolm: Moved to refactoring issue.

include/ghex/communication_object_ipr.hpp
bettiolm: Moved to refactoring issue.

include/ghex/communication_object_ipr.hpp
bettiolm: I changed the code a little bit here now. `field_info` is actually used only for the send part, so only with pack function. I added a specific struct for the receive, without callback (unpack) function.

bettiolm:
I addressed the comments and added a couple of issues (mostly minor refactoring) which can be addresses after the merge in my opinion. From my side it is ok now.

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -376,7 +386,7 @@ namespace gridtools {
                              typename Memory,
                              typename Field,
                              typename O>
-                    void allocate(Memory& mem,
+                    void allocate(Memory& mem, // reference not strictly needed here
```
mbianco, line 389: why is this comment useful? could you expand it?

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -410,23 +417,17 @@ namespace gridtools {
...
                              typename ValueType,
-                             typename BufferType,
                              typename Memory,
                              typename Halos,
-                             typename Function,
-                             typename DeviceIdType,
-                             typename Pool,
                              typename Field = void>
-                    void set_recv_size(Memory& memory,
+                    void set_recv_size(Memory& memory, // reference not strictly needed here
```
mbianco, line 423: why is this comment useful? could you expand it?

include/ghex/unstructured/communication_object_ipr.hpp
```diff
@@ -473,14 +474,13 @@ namespace gridtools {
...
                              typename ValueType,
-                             typename BufferType,
                              typename Memory,
                              typename Halos,
                              typename Function,
                              typename DeviceIdType,
                              typename Pool,
                              typename Field = void>
-                    void allocate_send(Memory& memory,
+                    void allocate_send(Memory& memory, // reference not strictly needed here
```
mbianco, line 483: why is this comment useful? could you expand it?

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: It's just a reminder, memory is already a pointer, and it is used as such inside, so a copy could also be ok. Same happens in comm2. I just did not want to change much the interface with respect of the comm2 one, since that should be harmless.

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: as before

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: as before

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: I pushed some documentation and removed the pass by reference from this function (allocate), in which it is actually not needed.

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Actually not, sorry. Here memory is not a pointer as before, and represents the recv memory. Reference is needed. I pushed the updated version.

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: Same as for `set_recv_size`

include/ghex/unstructured/communication_object_ipr.hpp
bettiolm: From my side it should be ok now

mbianco (approved):

PR #107, opened 2020-09-23
Multithreading tests for unstructured

bettiolm:


tests/unstructured_user_concepts.cpp
```diff
@@ -797,4 +800,94 @@ TEST(unstructured_user_concepts, data_descriptor_oversubscribe_ipr) {
 
 }
 
-#endif
+#else
+
+/** @brief Test data descriptor concept with multiple threads*/
```
mbianco, line 805: Are the tests self contained or are they using a common infrastructure? It may be useful to split the big file into smaller ones, I think.

bettiolm: They are all using a bunch of helper functions (to initialize data, check results, ...), but they can be easily included and the tests splitted, yep

mbianco: was this done?

bettiolm: Sorry, I left it half done. I've pushed everything now

include/ghex/unstructured/util/test_util.hpp
```diff
@@ -0,0 +1,439 @@
+﻿/*
```
mbianco, line 1: This file contains instances of grids, right? this could be better placed in `tests` folder?

bettiolm: I'd better move it there, yes. Also, it makes use of the GTEST facilities. Should I use a dedicated namespace for it?

mbianco: maybe, up to you. As the file is only used by one test file the namespace is not really necessary, I would say

bettiolm: Ok, done.

mbianco (approved):

PR #108, opened 2020-09-24
added exchange test

boeschf:
a simple additional test which exchanges with and without bulk/rma, and uses threads.


tests/simple_regular_exchange.cpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+TEST(simple_regular_exchange, single)
+{
+    sim(false);
+}
+
+TEST(simple_regular_exchange, multi)
+{
+    sim(false);
```
mbianco, line 241:
```suggestion
    sim(true);
```
?

boeschf: obviously! I will also add the gpu test now

include/ghex/bulk_communication_object.hpp
```diff
@@ -354,27 +384,41 @@ class bulk_communication_object
...
                 {
+                    int q=0;
                     for (auto it = h_it->second.rbegin(); it != h_it->second.rend(); ++it)
                     {
                         const auto& c = *it;
                         s_range.m_ranges.back().emplace_back(
-                            m_comm, f, c, h_it->first.mpi_rank, h_it->first.tag); 
+                            m_comm, f, c, h_it->first.mpi_rank
+                            , (m_it->second + h_it->first.tag+1)*10000 + q
+                            // alternatively rely on message ordering:
```
mbianco, line 398: If this is useful in some cases, shouldn't we guard this with a macro?

include/ghex/bulk_communication_object.hpp
```diff
@@ -354,27 +384,41 @@ class bulk_communication_object
...
+                    int q=0;
                     for (auto it = h_it->second.rbegin(); it != h_it->second.rend(); ++it)
                     {
                         const auto local = rma::is_local(m_comm, h_it->first.mpi_rank);
                         const auto& c = *it;
                         t_range.m_ranges.back().emplace_back(
-                            m_comm, f, field_info, c, h_it->first.mpi_rank, h_it->first.tag, local); 
+                            m_comm, f, field_info, c, h_it->first.mpi_rank
+                            , (m_it->second + h_it->first.tag+1)*10000 + q
+                            // alternatively rely on message ordering:
```
mbianco, line 415: as above

include/ghex/common/await_futures.hpp
```diff
@@ -12,75 +12,71 @@
...
-                }
-            }
-        }
+                        if (range[idx].test())
+                        {
+                            cont(range[idx].get());
+                            return true;
+                        } else return false;
+                    })
+                - index_list.begin());
```
mbianco, line 41: why do you need to resize all the time?

include/ghex/communication_object_2.hpp
```diff
@@ -22,6 +22,10 @@
 #include <stdio.h>
 #include <functional>
 
+// define this macro for fat callbacks
+// may not yet work optimally with the current ucx implementation
+//#define GHEX_COMM_OBJ_USE_FAT_CALLBACKS
```
mbianco, line 27: Where are these FAT call backs described? what are they?

mbianco (commented):
some comments

include/ghex/bulk_communication_object.hpp
boeschf: Maybe. I will add a macro.

include/ghex/common/await_futures.hpp
boeschf: the `index_list` is a helper data structure - it needs to be the same size as the vector of futures/request. Could you elaborate on your concerns here? The reason I made these helper vectors `static thread_local` is to avoid frequent reallocation (the `resize` will not trigger an allocation if the capacity is big enough).

include/ghex/communication_object_2.hpp
boeschf: I will extend this explanation

include/ghex/common/await_futures.hpp
mbianco: it's probably not a problem, but I would have done in in two phases: one I compute the number of things to remove and then I resize. Which is probably not better, since the resize if a decrement of a variable...

boeschf: ok, now I understood. Good point, have changed it accordingly.

include/ghex/communication_object_2.hpp
boeschf: ok, now I understood. Good point, have changed it accordingly.

include/ghex/common/await_futures.hpp
boeschf: ok, now I understood. Good point, have changed it accordingly.

include/ghex/structured/rma_put.hpp
```diff
@@ -39,8 +39,11 @@ using gpu_to_gpu = std::integral_constant<bool,
     std::is_same<typename SourceField::arch_type, gridtools::ghex::gpu>::value &&
     std::is_same<typename TargetField::arch_type, gridtools::ghex::gpu>::value>;
 
+// attributes needed for gcc to produce optimized code
 template<typename SourceField, typename TargetField>
-inline std::enable_if_t<
+__attribute__ ((optimize ("no-tree-loop-distribute-patterns")))
```
mbianco, line 44: should these be guarded with `__GNUC__` (I think this is the macro...)

include/ghex/rma/xpmem/access_guard.hpp
```diff
@@ -0,0 +1,136 @@
...
+    unsigned char volatile* get_ptr()
+    {
+        return (unsigned char volatile*)(m_handle->get_ptr());
+    } 
+
+    void start_source_epoch()
+    {
+        while(static_cast<unsigned char>(access_mode::remote) != get_ptr()[0])
+        {
+            // TODO call comm.progress()
```
mbianco, line 114: forgot about it?

mbianco (commented):
I added another couple of comments. I need to look at the domain decomposition and a couple of other things. Maybe tomorrow we can have a chat about it to speed up the process?

include/ghex/structured/rma_put.hpp
boeschf: yes, of course. Will add

mbianco (approved):

boeschf:
@mbianco thanks for approving. I refactored the decomposition so it can be more easily used with different number of hierarchy levels (this will be helpful when for example also considering L3 cache). I also extended the unit test for the decomposition. Let me know if I can go ahead and merge.

include/ghex/util/resource_layout.hpp
```diff
@@ -0,0 +1,170 @@
...
+        static_assert(I < Levels, "level index I must be within the range of specified levels");
+        return m_dist(idx)[I];
+    }
+
+    /** returns resource index relative to level I, given a global resource index */
+    template<unsigned int I>
+    size_type relative_resource(size_type idx) const noexcept
+    {
+        static_assert(I < (Levels-1), "level I must be lower than greatest level");
+        //return relative_resource<I-1>(idx) - index<I>(idx)*m_1D_dist[I].size();
```
boeschf, line 117:
```suggestion
```

PR #117, opened 2021-03-23
new ucx version

boeschf:
upgrade to new ucx version

### DIFF
--- a/include/ghex/transport_layer/ucx/endpoint.hpp
+++ b/include/ghex/transport_layer/ucx/endpoint.hpp
@@ -64,7 +64,7 @@ namespace gridtools {
                     void destroy()
                     {
                         if (!m_moved){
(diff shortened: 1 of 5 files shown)


include/ghex/transport_layer/ucx/endpoint.hpp
```diff
@@ -62,7 +64,7 @@ namespace gridtools {
                     void destroy()
                     {
                         if (!m_moved){
-                            ucs_status_ptr_t ret = ucp_ep_close_nb(m_ep, UCP_EP_CLOSE_MODE_FLUSH);
```
angainor, line 65: I am a bit unsure about this one. I think we should leave `UCP_EP_CLOSE_MODE_FLUSH`, at least this is what the UCX guys advised. This might require a synchronization before EPs are destroyed (so a barrier before destructors are called) to avoid a situation where the peer worker has already been destroyed when we call `ucp_ep_close_nb`).

angainor (approved):
The UCX changes are fine - they simply add support for ANY source, nothing else.

PR #120, opened 2021-04-10
Porting to AMD GPUs

bettiolm:


bettiolm:
I just added a new Dockerfile for the CPU tests, with a predefined ucx version (1.10.1) and openmpi built from source instead of using ubuntu binaries. This does not have a direct impact on the CI, but could be used to build and upload a new image to pull from.

bettiolm:
Note on the CI: current images on Docker Hub have been fully replaced with new images built using the `gridtools-docker` repository, on which specific layers for GHEX have been added. The new environments are exactly the same as the previous ones, except for the `CXX` environment variable not being set to `clang++` in the amd gpu image (thus requiring to set it explicitly in the configure command of the CI workflow).

PR #125, opened 2021-04-28
Unstructured benchmark using ParMeTiS graph partitioning

bettiolm:


mbianco:
Remove the pictures form the copy

bettiolm:
> Remove the pictures form the copy

Removed and modified GHEX wiki accordingly

benchmarks/CMakeLists.txt
```diff
@@ -316,6 +316,43 @@ endforeach()
 add_subdirectory(transport)
 
 
+# unstructured benchmarks
+# ----------------
+
+if (GHEX_ENABLE_PARMETIS_BINDINGS)
```
mbianco, line 322: The comment below suggest that the variable name should be changed into `GHEX_UNSTRUCTURED_MT` or something?

benchmarks/unstructured_parmetis.cpp
```diff
@@ -0,0 +1,793 @@
+﻿/*
```
mbianco, line 1: This is another big file with some replication... Would it be better to split it into individual programs, or maybe reuse more code?

benchmarks/CMakeLists.txt
bettiolm: I see, I changed the comment a bit. In the following foreach, CPU benchmarks support multithreading and the actual number of threads is controlled by an environment variable (single-thread is also possible), whereas the GPU one is single-threaded. I would say the common thing among this set of benchmarks is that they all need ParMeTis bindings, whether threading mode may vary.

benchmarks/unstructured_parmetis.cpp
bettiolm: I can do something here, sure. One thing could be to wrap the parmetis repartitioning section into something that might be useful for further tests / benchmarks. Also, I was thinking of generalizing e.g. the reindexing for the in-place receive in a future PR for the new Atlas bindings, since it could be also useful in general.

bettiolm:
From my side, this PR seems OK now

PR #129, opened 2021-06-17
Communication object 2 test suite, multiple GPUs per node + threads: setting device

bettiolm:


tests/communication_object_2.cpp
```diff
@@ -459,8 +459,11 @@ TEST(communication_object_2, exchange)
 #endif
 
 #ifdef GHEX_TEST_THREADS
-    auto func = [&context](auto... bis)
+    auto func = [&context, &local_comm](auto... bis)
```
mbianco, line 462: Maybe instead of passing the MPI communicator, the thread could take a `local_rank` integer argument (by value then)?
It's unfortunate that we have the `#ifdef` in the thread body, but I do not see how else to do this...

bettiolm: Changed it, now I pass the local rank by value.
On the second point, yes, I agree. I might wrap the `setDevice(...)` function maybe, but it won't change substantially.

mbianco (approved):

PR #132, opened 2021-07-02
remap MPI ranks to PMIx ranks in the PMIx address db

angainor:
This is necessary in cases when ghex is initialized with a communicator with remapped ranks (e.g., hwcart). In that case PMIx has the same numbering as MPI_COMM_WORLD, but GHEX is using different rank numbers that come from the renumbered communicator. To use PMIx those need to be remapped.

### DIFF
--- a/include/ghex/transport_layer/ucx/address_db_pmi.hpp
+++ b/include/ghex/transport_layer/ucx/address_db_pmi.hpp
@@ -37,6 +37,7 @@ namespace gridtools {
                     using value_t   = address_t;
 
                     MPI_Comm m_mpi_comm;
+                    std::vector<key_t> m_rank_map;
 
                     // these should be PMIx ranks. might need remaping to MPI ranks
                     key_t m_rank;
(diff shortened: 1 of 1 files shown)


boeschf (approved):

include/ghex/transport_layer/ucx/address_db_pmi.hpp
```diff
@@ -63,10 +64,18 @@ namespace gridtools {
...
+                            throw std::runtime_error("PMIx and MPI sizes are different. Bailing out.");
+                        }
 
-                        // TODO: m_mpi_comm should be use to obtain MPI rank to PMIx rank map
-                        if(m_rank != mpi_rank || m_size != mpi_size)
-                            throw std::runtime_error("PMIx and MPI ranks are different. Mapping not implemented yet.");
+                        // map MPI communicator ranks to PMIx ranks
+                        m_rank_map.resize(mpi_size);
+                        for(int i=0; i<mpi_size; i++) m_rank_map[i] = m_rank;
+                        GHEX_CHECK_MPI_RESULT(MPI_Alltoall(MPI_IN_PLACE, 0, MPI_INT, m_rank_map.data(), sizeof(key_t), MPI_BYTE, comm));
```
boeschf, line 75:
```suggestion
                        GHEX_CHECK_MPI_RESULT(MPI_Alltoall(MPI_IN_PLACE, 0, MPI_BYTE, m_rank_map.data(), sizeof(key_t), MPI_BYTE, comm));
```

PR #133, opened 2021-07-02
fixes in fortran bindings

angainor:
- convert fortran logicals (periodic) to C ints
 - fix callback function signature

### DIFF
--- a/tests/bindings/test_f_send_recv_cb.f90
+++ b/tests/bindings/test_f_send_recv_cb.f90
@@ -122,9 +122,12 @@ contains
     type(ghex_cb_user_data), value :: user_data
     integer :: thrid
     integer(1), dimension(:), pointer :: msg_data
-    procedure(f_callback), pointer :: pcb
     integer, pointer :: received
-    pcb => recv_callback
+
+    ! NOTE: this segfaults in Intel compiler. It seems we have to use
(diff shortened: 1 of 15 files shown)


boeschf (approved):

tests/bindings/test_f_send_recv_cb.f90
```diff
@@ -122,9 +122,12 @@ subroutine recv_callback (mesg, rank, tag, user_data) bind(c)
     type(ghex_cb_user_data), value :: user_data
     integer :: thrid
     integer(1), dimension(:), pointer :: msg_data
-    procedure(f_callback), pointer :: pcb
     integer, pointer :: received
-    pcb => recv_callback
+
+    ! NOTE: this segfaults in Intel compiler. It seems we have to use
+    ! the globally defined pcb from the main function. WHY??
```
bettiolm, line 128: Is this question still open?

bettiolm (approved):

tests/bindings/test_f_send_recv_cb.f90
angainor: @bettiolm yes, unfortunately.The current code works, but I would expect the commented code to also work. It does not work on Intel, and I don't yet know why.

PR #134, opened 2021-07-04
Python bindings

tehrengruber:
This is a first proof-of-concept implementation of the Python bindings that still needs some cleanup and polishing. The general idea is that for each class in ghex there exists a cpp file with the bindings and the same base path in `bindings/python/cpp/src`. To speed up compilation each file is compiled in a seperate translation unit.

Readme: https://github.com/tehrengruber/GHEX/blob/py_bindings/bindings/python/README.md
Test case: https://github.com/tehrengruber/GHEX/blob/py_bindings/bindings/python/tests/test_mpi4py.py


bindings/python/cpp/src/ghex.cpp
```diff
@@ -0,0 +1,130 @@
...
+    using pattern_type = decltype(gridtools::ghex::make_pattern<GridType>(std::declval<context_type&>(), std::declval<HaloGenerator&>(), std::declval<std::vector<Domain>&>()));
+    using communicator_type = typename context_type::communicator_type;
+    using communication_object_type = gridtools::ghex::communication_object<communicator_type, grid_type, domain_id_type>;
+
+
+    // todo: just for getting something runnable
+    using arch = gridtools::ghex::cpu;
+    using domain_descriptor_type = gridtools::ghex::structured::regular::domain_descriptor<domain_id_type, std::integral_constant<int, 3>>;
+    using array_type = std::array<int, 3>;
+    auto wrapper = [] (const domain_descriptor_type& dom, double* data, const array_type& offsets, const array_type& extents) {
```
bettiolm, line 87: Maybe some code reuse is possible here?

bindings/python/cpp/src/ghex.cpp
```diff
@@ -0,0 +1,130 @@
...
+        return gridtools::ghex::wrap_field<arch, ::gridtools::layout_map<0,1,2>>(dom, data, offsets, extents);
+    };
+    using field_t = decltype(wrapper(std::declval<const domain_descriptor_type&>(), std::declval<double*>(), std::declval<const array_type&>(), std::declval<const array_type&>()));
+
+    using buffer_info_type = decltype(std::declval<pattern_type>()(std::declval<field_t&>()));
+
+    py::class_<communication_object_type>(m, "CommunicationObject")
+        .def(py::init([] (communicator_type c) {
+            return gridtools::ghex::make_communication_object<pattern_type>(c); }))
```
bettiolm, line 96: In principle, could a similar approach (make_communication_object... called within python init function) be used also before, for the pattern?

bindings/python/cpp/src/ghex.cpp
```diff
@@ -0,0 +1,130 @@
...
+        .def("exchange", [] (communication_object_type& co, buffer_info_type& b) { return co.exchange(b); });
```
bettiolm, line 97: Why using a lambda here, and not just bind to &communication_object::exchange? Is it because it is simpler than using the overloaded methods syntax?

bettiolm (commented):
First interaction, mostly minor comments and personal notes.

boeschf:
closing this as it has been updated and incorporated into #146

PR #145, opened 2022-08-02
Support for GT2.2

fthaler:
GridTools 2.2 removes `gridtools/meta/type_traits.hpp`, as this was providing ports to C++14 for some traits introduced to the STL with C++17 (which is a requirement for GridTools version ≥ 2.2). With this PR, GHEX supports both GridTools versions < 2.2 with C++14 (or C++17) and GridTools versions ≥ 2.2 with C++17.

### DIFF
--- a/include/ghex/glue/gridtools/field.hpp
+++ b/include/ghex/glue/gridtools/field.hpp
@@ -10,12 +10,13 @@
  */
 #pragma once
 
+#include <type_traits>
+
 #include <ghex/config.hpp>
 #include <ghex/arch_traits.hpp>
 #include <ghex/structured/regular/field_descriptor.hpp>
 #include <ghex/glue/gridtools/processor_grid.hpp>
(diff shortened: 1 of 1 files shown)


bettiolm:
Sure, that's ok for me. Only thing is that we should probably also change `set(CMAKE_CXX_STANDARD 14)` to `17` in the main `CMakeLists.txt`.

fthaler:
I don’t know if you want to introduce that change with this PR. This is just a compatibility fix for GT 2.2 (which only supports C++ 17), but GHEX still supports GT ≥ 2.0 with C++14. But of course I could add that…

bettiolm:
Yep, agree. Sure, Let's keep the CMake part as it is then, maybe we will change it later. From my side, we can just merge the PR.

boeschf (approved):

fthaler:
Thanks! Could you please merge it then, I don’t have write access :)

PR #146, opened 2023-05-02
Refactoring2

boeschf:
- update ghex to latest oomph
- add some long outstanding stuff
- fortran bindings for unstructured
- python bindings


CMakeLists.txt
```diff
@@ -1,26 +1,66 @@
...
+    # Enable CDash dashboard testing integration
+    include(CTest)
+endif()
+
+
+# ---------------------------------------------------------------------
+# Dependencies
+# ---------------------------------------------------------------------
+include(ghex_external_dependencies)
+target_link_libraries(ghex_common INTERFACE MPI::MPI_CXX)
```
mbianco, line 52: Are we sending on the C++ interfaces to MPI? As far as I know they have been deprecated years ago...

bindings/fhex/ghex_unstructured_mod.f90
```diff
@@ -0,0 +1,221 @@
...
+! ghex-org
+!
+! Copyright (c) 2014-2023, ETH Zurich
+! All rights reserved.
+!
+! Please, refer to the LICENSE file in the root directory.
+! SPDX-License-Identifier: BSD-3-Clause
+!
+MODULE ghex_unstructured_mod
+    !
```
mbianco, line 11: Why is the copyright inside the module?

bindings/fhex/ghex_unstructured_mod.f90
```diff
@@ -0,0 +1,221 @@
+!
+! ghex-org
+!
+! Copyright (c) 2014-2023, ETH Zurich
+! All rights reserved.
+!
+! Please, refer to the LICENSE file in the root directory.
+! SPDX-License-Identifier: BSD-3-Clause
+!
+MODULE ghex_unstructured_mod
```
mbianco, line 10: Ideally we have some documentation on how to use this... I hope the test will show how to use it...

bindings/python/mpi_comm_shim.hpp
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include <mpi.h>
+#include <pybind11/pybind11.h>
+#include <util/to_string.hpp>
+
+namespace pyghex
+{
+
+// https://github.com/arbor-sim/arbor/commit/1d6a48d0ce4b96f59acf931efd61d55c571c5e68
```
mbianco, line 19: Why a reference to arbor?

bindings/python/structured/grid.py
```diff
@@ -0,0 +1,321 @@
+#
```
mbianco, line 1: This seems to be an implementation in Python of the Grid, and next there's also index_space. Why do you need a python implementation instead of challenging to the C++ implementation?

bindings/python/util/demangle.hpp
```diff
@@ -0,0 +1,36 @@
...
+namespace util
+{
+
+template<typename T>
+std::string
+demangle()
+{
+    int                                    status = 0;
+    std::unique_ptr<char, void (*)(void*)> res{
+        abi::__cxa_demangle(typeid(T).name(), nullptr, nullptr, &status), std::free};
```
mbianco, line 29: What compilers support this? I think clang does, right? But I'm not sure about intel or Cray. This may limit the support of GHEX for some compilers. Fine for now, but we may document this in the readme?

include/ghex/communication_object.hpp
```diff
@@ -365,15 +364,15 @@ class communication_object
         using test_t = pattern_container<grid_type, domain_id_type>;
         std::map<const test_t*, int> pat_ptr_map;
         int                          max_tag = 0;
-        for_each(iter_pairs_t, [&pat_ptr_map, &max_tag](auto iter_pair) {
+        for_each(iter_pairs_t, [&pat_ptr_map, &max_tag](std::size_t, auto iter_pair) {
```
mbianco, line 367: What is happening here, why is a size_t appearing in the lambda, but not used?

include/ghex/mpi/communicator.hpp
```diff
@@ -191,6 +216,133 @@ class communicator
...
+        const T*    _begin;
+        const T*    _end;
+        const T*    begin() const noexcept { return _begin; }
+        const T*    end() const noexcept { return _end; }
+        std::size_t size() const noexcept { return _end - _begin; }
+    };
+
+    // This algorithm is a slight adaptation of Simon Frasch's implementation here:
+    // https://github.com/AdhocMan/arbor/commit/75fa66ab983598c1c7f5e064cde4a062e9d030b5#diff-a5a30d5a5b994ff68317d61e212c8d959c209b2fa4caa3f9cd2e738b9e55ef63
+    template<typename F, typename... Ts>
```
mbianco, line 233: This also seems to come from arbor! good. Usually in algorithm implementaiton the functor is specified after the data. Is there a reason why here it is inverted? (no need to change it, just curiosity)

mbianco: What is the purpose of this? In my understanding a distributed for each is simply running the loop on the differetn ranks with the local sizes. Why do you need reductions? I did not find uses for this algorithm in this PR (but it is huge, my computer is collapsing :) )

mbianco (commented):
Did go over half of the files, maybe. I have just few questions, but nothing blocking.

include/ghex/communication_object.hpp
boeschf: The for each was changed to also pass an iteration count as first argument to the user supplied visitor function.

include/ghex/mpi/communicator.hpp
boeschf: - Regarding the first point: that's a limitation of current c++ where the parameter pack in a function must appear last in order to be able to deduce the other types (here `F`).
- The purpose: it is an alternative way to do an `allgather_v` with multiple vectors at once. The difference to a conventional allgather_v is that we never store all data but rather copy over each rank's data one at a time. For large simulations the local memory might simply be too small to store all global data. For now, it is used in the [creation of the unstructured pattern](https://github.com/boeschf/GHEX/blob/612bf7eb6b59310a100b8a942c911889d076a525/include/ghex/unstructured/pattern.hpp#L188). I will probably also apply this algorithm in the structured pattern in a later PR.

bindings/fhex/ghex_unstructured_mod.f90
boeschf: that's an oversight, will remove

CMakeLists.txt
boeschf: No, we use the C interface. This seems just to be the way to use the cmake module.

bindings/python/mpi_comm_shim.hpp
boeschf: taken from Arbor, will add clarification

bindings/python/structured/grid.py
tehrengruber: The Grid implementation here is completely superfluous and removed now. The Cartesian Sets are a neat way to work with domains and provide a nice interface to the user that goes beyond what the C++ implementation provides. To be fair they have far more features than strictly needed though. For example the IndexSpace is never used in the bindings themselves, but make the tests very short as a domain decomposition becomes just a single line. For now I would keep this and suggest we revisit this again when we have a better picture on how people use the bindings in their codes.

PR #158, opened 2024-03-11
Expose C++ obj pointers to Python API

kotsaloscv:
Th exposed pointers are passed to [DaCe](https://github.com/spcl/dace) which handles the halo exchange through GHEX.

### DIFF
--- a/bindings/python/module.cpp
+++ b/bindings/python/module.cpp
@@ -7,7 +7,6 @@
  * Please, refer to the LICENSE file in the root directory.
  * SPDX-License-Identifier: BSD-3-Clause
  */
-#include <cstdint>
 #include <pybind11/pybind11.h>
 #include <gridtools/common/for_each.hpp>
 #include <context_shim.hpp>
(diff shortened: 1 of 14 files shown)


bindings/python/module.cpp
```diff
@@ -72,4 +77,36 @@ PYBIND11_MODULE(_pyghex, m)
     pyghex::unstructured::register_field_descriptor(m);
     pyghex::unstructured::register_pattern(m);
     pyghex::unstructured::register_communication_object(m);
+
+    m.def("expose_cpp_ptr", [](pyghex::context_shim* obj) {return reinterpret_cast<uintptr_t>(&obj->m);} );
```
boeschf, line 81: please `#include <cstdint>` and
```suggestion
    m.def("expose_cpp_ptr", [](pyghex::context_shim* obj) {return reinterpret_cast<std::uintptr_t>(&obj->m);} );
```

bindings/python/module.cpp
```diff
@@ -72,4 +77,36 @@ PYBIND11_MODULE(_pyghex, m)
...
+            m.def("expose_cpp_ptr", [](pattern_container* obj) {return reinterpret_cast<uintptr_t>(obj);} );
+        });
+
+    gridtools::for_each<
+        gridtools::meta::transform<gridtools::meta::list, pyghex::unstructured::domain_descriptor_specializations>>(
+        [&m](auto l)
+        {
+            using type = gridtools::meta::first<decltype(l)>;
+            m.def("expose_cpp_ptr", [](type* obj) {return reinterpret_cast<uintptr_t>(obj);} );
+        });
```
boeschf, line 111: I would move those functions into their respective source files to have only one loop over the types.

kotsaloscv:
replace by #163

PR #161, opened 2024-03-26
spack ci

boeschf:
- add spack to to ci
- update readme


msimberg:
I'll close this since we have CSCS CI set up with spack builds now.

PR #170, opened 2024-12-13
PyPI deploy action

havogt:


bindings/python/README.md
```diff
@@ -0,0 +1,41 @@
+[![License](https://img.shields.io/badge/License-BSD%203--Clause-blue.svg)](https://opensource.org/licenses/BSD-3-Clause)
```
havogt, line 1: I removed the C++ specific stuff from the root/README, not sure if a good idea, or just linking to the root one is good enough. It's the page that is displayed here https://test.pypi.org/project/ghex/.

boeschf (approved):

PR #178, opened 2025-01-04
Add support for bool, int32, int64 in python bindings

tehrengruber:


bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
```diff
@@ -0,0 +1,50 @@
+/*
```
tehrengruber, line 1: Not really happy about the name and a new file. I'm open for suggestions, but also ok with this.

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
```diff
@@ -0,0 +1,50 @@
...
+        if (dtype.is(py::dtype::of<type>())) {
+            assert(cpp_name.empty());
+            cpp_name = util::demangle<type>();
+        }
+    });
+
+    if (cpp_name.empty()) {
+        throw std::invalid_argument("Unsupported numpy dtype");
+    }
+
```
havogt, line 40: maybe better to handle the error at the call site (unless exceptions in C++ are commonly used in the bindings...)

tehrengruber:
The container images are out-dated.
- gcc-9-ucx-mpi-atlas-parmetis-hip: not built anymore. Removed in https://github.com/GridTools/gridtools-docker/pull/96/files
- gcc-9-ucx-mpi-atlas-parmetis-cuda-11: The naming scheme has changed the minor version is now part of the name, i.e. it should be `gcc-9-ucx-mpi-atlas-parmetis-cuda-11.8`

havogt:
What do you mean by out-dated? Do you need a more recent compiler or is the image not compatible with the ubunut 24.04 runner?

tehrengruber:
Ghex uses container images that are not built by https://github.com/GridTools/gridtools-docker anymore. The `gcc-9-ucx-mpi-atlas-parmetis-cuda-11` is easy to fix, not sure what about `gcc-9-ucx-mpi-atlas-parmetis` though. The issue is that those images don't use ubuntu 24.04, but 22.04. The pybind11 package that comes with that version is to old to provide `item_type_is_equivalent_to` which is used in this PR (https://github.com/pybind/pybind11/commit/8e1f9d5c40f74632799c8a44287d32d18a6c8630).

boeschf (approved):

PR #198, opened 2026-03-19
Migrate from pybind11 to nanobind

msimberg:


bindings/python/src/_pyghex/unstructured/communication_object.cpp
```diff
@@ -42,55 +46,56 @@ extract_cuda_stream(pybind11::object python_stream)
...
-            const auto protocol_version = cuda_stream_protocol[0].cast<std::size_t>();
+            const auto protocol_version = nanobind::cast<std::size_t>(cuda_stream_protocol[0]);
             if (protocol_version == 0)
             {
                 std::stringstream error;
                 error << "Expected `__cuda_stream__` protocol version 0, but got "
                       << protocol_version;
-                throw pybind11::type_error(error.str());
+                throw nanobind::type_error(error.str().c_str());
             }
```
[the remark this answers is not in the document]
msimberg: #202 was merged, and I've updated this PR to include the fix.

bindings/python/src/_pyghex/mpi_comm_shim.cpp
```diff
@@ -30,11 +32,11 @@ namespace
...
 inline MPI_Comm
-convert_to_mpi_comm(pybind11::object o)
+convert_to_mpi_comm(nanobind::object o)
 {
-    if (!pybind11::hasattr(o, "py2f"))
-        throw pybind11::type_error("Argument must be `mpi4py.MPI.Comm`");
-    return MPI_Comm_f2c(o.attr("py2f")().cast<MPI_Fint>());
+    if (!nanobind::hasattr(o, "py2f"))
+        throw nanobind::type_error("Argument must be `mpi4py.MPI.Comm`");
+    return MPI_Comm_f2c(nanobind::cast<MPI_Fint>(o.attr("py2f")()));
```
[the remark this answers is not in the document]
msimberg: @copilot double check this. `o.attr("py2f")()` _does_ call the method, right? Perhaps counting parentheses is not your strength (or it's not mine)?

bindings/python/src/ghex/CMakeLists.txt
```diff
@@ -8,10 +8,13 @@ else()
...
-    add_dependencies(pyghex pyghex_files)
+    file(GLOB_RECURSE pyghex_python_files CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/*.py")
+
+    add_custom_target(pyghex_files ALL
+        COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR} "${CMAKE_CURRENT_BINARY_DIR}/../../ghex"
+        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/version.txt "${CMAKE_CURRENT_BINARY_DIR}/../../ghex"
+        DEPENDS ${pyghex_python_files} ${CMAKE_BINARY_DIR}/version.txt
+        COMMENT "Refreshing Python binding files for tests")
+
+    add_dependencies(pyghex_files pyghex)
```
msimberg, line 19: This change is meant to ensure that python files are reinstalled when they change. It's not strictly pybind11 to nanobind related, but it's useful to fix this when testing the python bindings.

bindings/python/src/_pyghex/unstructured/field_descriptor.cpp
```diff
@@ -19,97 +19,47 @@
 #include <register_class.hpp>
```
msimberg, line 19: This file and the `regular/field_descriptor.cpp` files needed the biggest changes and may need some extra care when reviewing.

msimberg:
I've merged #208 into this branch, so python tests should now also run on CSCS CI pipelines.

.cscs-ci/container/deps.Containerfile
```diff
@@ -8,8 +8,9 @@ RUN mkdir -p /opt/spack && \
 ENV PATH="/opt/spack/bin:$PATH"
 
 ARG SPACK_PACKAGES_SHA
+# TODO: Change msimberg back to upstream
 RUN mkdir -p /opt/spack-packages && \
-    curl -fLsS "https://api.github.com/repos/spack/spack-packages/tarball/$SPACK_PACKAGES_SHA" | tar --strip-components=1 -xz -C /opt/spack-packages
+    curl -fLsS "https://api.github.com/repos/msimberg/spack-packages/tarball/$SPACK_PACKAGES_SHA" | tar --strip-components=1 -xz -C /opt/spack-packages
```
msimberg: To do: this and the SPACK_PACKAGES_SHA need to be changed back to upstream before merging.

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
```diff
@@ -36,90 +41,44 @@ template<int... val>
...
-    }
 };
 
-template<typename Arch>
-pybind11::buffer_info
-get_buffer_info(pybind11::object& buffer)
+template<typename T>
+std::vector<std::ptrdiff_t>
+byte_strides(const T& b, std::ptrdiff_t itemsize)
 {
```
philip-paul-mueller:
```suggestion
template<typename... Args>
std::vector<std::ptrdiff_t>
byte_strides(const nb::ndarray<Args...>& b)
{
    const std::ptrdiff_t itemsize = b.itemsize();
```
Maybe a bit cleaner.

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
```diff
@@ -135,43 +94,54 @@ register_field_descriptor(pybind11::module& m)
                 ghex::buffer_info<pattern_type, arch_type, field_descriptor_type>;
 
             auto _field_descriptor = register_class<field_descriptor_type>(m);
-            /*auto _buffer_info =*/register_class<buffer_info_type>(m);
+            register_class<buffer_info_type>(m);
```
philip-paul-mueller:
```suggestion
            (void)register_class<buffer_info_type>(m);
```

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
```diff
@@ -135,43 +94,54 @@ register_field_descriptor(pybind11::module& m)
...
+            {
+                if (b.ndim() != dimension::value)
+                {
+                    std::stringstream error;
+                    error << "Field has wrong dimensions. Expected " << dimension::value
+                          << ", but got " << b.ndim();
+                    throw nanobind::type_error(error.str().c_str());
+                }
+
+                auto strides = byte_strides(b, sizeof(T));
```
philip-paul-mueller:
```suggestion
                auto strides = byte_strides(b);
```

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
```diff
@@ -135,43 +94,54 @@ register_field_descriptor(pybind11::module& m)
...
+
+                auto ordered_strides = strides;
```
philip-paul-mueller, line 115: As far as I can see from the use bellow you kind of assume that the strides are unique.
While I think this is the case most of the times I am not sure if it is always true.
I think this is only true if all dimensions have a different size and the dimension of size 1 is at one of the ends.

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
```diff
@@ -135,43 +94,54 @@ register_field_descriptor(pybind11::module& m)
...
+            _field_descriptor.def(
+                "__init__",
+                [make_field_descriptor](field_descriptor_type* t, const domain_descriptor_type& dom,
+                    nanobind::ndarray<T, typename ndarray_device_type<arch_type>::type> b,
+                    const array& offsets, const array& extents)
+                { new (t) field_descriptor_type(make_field_descriptor(dom, b, offsets, extents)); },
+                nanobind::keep_alive<1, 3>());
+#else
+            _field_descriptor.def(nanobind::new_(make_field_descriptor),
+                nanobind::keep_alive<0, 3>());
```
philip-paul-mueller, line 144: I am actually not sure if you must use `keep_alive<1, 3>()` here.
Because according to the [documentation](https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4I_6size_t_6size_tEN8nanobind10keep_aliveE) `1` "ndex 1 refers to the first argument. In methods and constructors, index 1 refers to the implicit this pointer" however, I am not sure if it applies here as well.

bindings/python/src/_pyghex/unstructured/communication_object.cpp
```diff
@@ -106,9 +110,9 @@ register_communication_object(pybind11::module& m)
 #if defined(GHEX_CUDACC)
                 .def(
                     "schedule_wait",
-                    [](typename type::handle_type& h, pybind11::object python_stream)
+                    [](typename type::handle_type& h, nanobind::object python_stream)
                     { return h.schedule_wait(extract_cuda_stream(python_stream)); },
-                    pybind11::keep_alive<0, 1>())
+                    nanobind::keep_alive<0, 1>(), nanobind::arg("stream").none())
```
philip-paul-mueller: Not part of the PR but this looks strange.
As far as I can tell/remember `schedule_wait()` does not return anything.
And according to the [documentation](https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4I_6size_t_6size_tEN8nanobind10keep_aliveE) it should not be used if the function does not return anything.
I am not fully sure if we should remove it.

bindings/python/src/_pyghex/unstructured/communication_object.cpp
```diff
@@ -124,52 +128,52 @@ register_communication_object(pybind11::module& m)
...
                             "schedule_exchange",
-                            [](type& co, pybind11::object python_stream,
+                            [](type& co, nanobind::object python_stream,
                                 std::vector<buffer_info_type> b) {
                                 return co.schedule_exchange(extract_cuda_stream(python_stream),
                                     b.begin(), b.end());
                             },
-                            pybind11::keep_alive<0, 1>(), pybind11::arg("stream"),
-                            pybind11::arg("patterns"))
+                            nanobind::keep_alive<0, 1>(), nanobind::arg("stream").none(),
```
philip-paul-mueller, line 151: Not sure if the `.none()` is legal here or has an effect.
As stated in https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4N8nanobind3arg4noneEb it can not be used for built in types.
I know it worked in PyBind, but not sure if it still works in nanobind.
Is there a test for that?

bindings/python/src/_pyghex/unstructured/field_descriptor.cpp
```diff
@@ -121,107 +71,94 @@ register_field_descriptor(pybind11::module& m)
...
+                return type{dom, static_cast<T*>(b.data()), levels, levels_first, outer_strides};
+            };
+
+#if NB_VERSION_MAJOR < 2
+            _field_descriptor.def(
+                "__init__",
+                [make_field_descriptor](type* t, const domain_descriptor_type& dom,
+                    nanobind::ndarray<T, typename ndarray_device_type<arch_type>::type> b)
+                { new (t) type(make_field_descriptor(dom, b)); },
+                nanobind::keep_alive<1, 3>());
```
philip-paul-mueller, line 158: Same comment as in the other file applies.

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
```diff
@@ -9,30 +9,59 @@
...
 
-namespace py = pybind11;
-
 namespace pyghex
 {
 
+namespace
+{
+nanobind::object
+numpy_dtype(const nanobind::handle& value)
```
philip-paul-mueller: I would always use `object` might be tiny bit slower but not affected by wrong reference counting.

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
```diff
@@ -9,30 +9,59 @@
...
+    else if constexpr (std::is_same_v<T, std::int64_t>)
+        return numpy_dtype(nanobind::str("int64"));
+    else
+        static_assert(sizeof(T) == 0, "unsupported dtype");
+}
+} // namespace
+
 std::string
-py_dtype_to_cpp_name(py::dtype dtype)
+py_dtype_to_cpp_name(nanobind::handle dtype)
```
philip-paul-mueller: I would also use an `object` here.

cmake/ghex_python.cmake
```diff
@@ -1,17 +1,14 @@
...
 
-    find_package (Python3 REQUIRED COMPONENTS Interpreter Development.Module)
+    find_package(Python 3 REQUIRED COMPONENTS Interpreter Development.Module)
 
-    if(${Python3_FOUND})
-        set(PYTHON_EXECUTABLE "${Python3_EXECUTABLE}")
-    endif()
+    set(PYTHON_EXECUTABLE "${Python_EXECUTABLE}")
 
     include(ghex_find_python_module)
```
philip-paul-mueller: There is still the error that makes `nanobind` an optional dependency despite a required one.
```
--- a/cmake/ghex_find_python_module.cmake
+++ b/cmake/ghex_find_python_module.cmake
@@ -5,7 +5,7 @@ function(find_python_module module)
 
     if(NOT PY_${module_upper})
         if(ARGC GREATER 1 AND ARGV1 STREQUAL "REQUIRED")
-            set(${module}_FIND_REQUIRED TRUE)
+           set(PY_${module}_FIND_REQUIRED TRUE)
         endif()
```

bindings/python/src/_pyghex/unstructured/communication_object.cpp
msimberg:
```suggestion
                    nanobind::arg("stream").none())
```

bindings/python/src/_pyghex/unstructured/communication_object.cpp
msimberg: The `none` is required, otherwise `None` will not be accepted as an argument. See https://nanobind.readthedocs.io/en/latest/functions.html#none-arguments for details.

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
```diff
@@ -9,30 +9,59 @@
...
 {
 
+namespace
+{
+nanobind::object
+numpy_dtype(const nanobind::handle& value)
+{
+    return nanobind::module_::import_("numpy").attr("dtype")(
+        nanobind::borrow<nanobind::object>(value));
+}
```
msimberg:
```suggestion
nanobind::object
numpy_dtype(char const* value)
{
    return nanobind::module_::import_("numpy").attr(value);
}
```

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
msimberg: Definitely. Not relying on itemsize was an overeager compatibility change... itemsize was added in 1.3.0, but there's no reason to support anything older than that (even supporting 1.X in general is a bit questionable, but nice for CI which uses very old versions of everything).

I'll change this.

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
msimberg: We'll investigate this separately. The logic is the same as before. A bit suspicious, but there might be some underlying assumptions that make this safe.

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
msimberg: Leaving this unchanged. Likely related to the construction being different. In the new case the buffer is kept alive with the lifetime of the return type (0) instead. In the `< 2` case nothing is directly returned.

bindings/python/src/_pyghex/unstructured/field_descriptor.cpp
msimberg: Keeping this one as well.

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
msimberg: I will check if we can avoid this whole dtype complication.

cmake/ghex_python.cmake
msimberg: I've cherry-picked your changes from #206.

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
msimberg: I simplified this heavily and now just rely on passing an ndarray instead from the python side, and access `nanobind::ndarray::dtype` on the c++ side.

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
msimberg: This function now directly takes a `nanobind::ndarray` instead.

bindings/python/src/_pyghex/structured/regular/field_descriptor.cpp
msimberg: Changed.

cmake/ghex_find_python_module.cmake
```diff
@@ -31,7 +31,7 @@ function(find_python_module module)
         else()
             set(HAVE_${module_upper} OFF CACHE INTERNAL "Python module available")
         endif()
-    endif(NOT PY_${module_upper})
+    endif()
```
msimberg:
```suggestion
    endif(NOT PY_${module_upper})
```

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
```diff
@@ -9,30 +9,29 @@
...
     std::string cpp_name;
 
     gridtools::for_each<pyghex::types::data>(
-        [&cpp_name, &dtype](auto l)
+        [&cpp_name, &array](auto l)
         {
             using type = decltype(l);
 
-            if (dtype.is(py::dtype::of<type>()))
+            if (array.dtype() == nanobind::dtype<type>())
```
msimberg, line 36: @philip-paul-mueller note that I changed this to a much simpler version. There doesn't seem to be an equivalent to pybind11's version of directly taking a dtype here, but taking an ndarray gives access to dtype (which nanobind has internally inferred from the dlpack interface). I think this is simpler than manually trying to reconstruct a dtype object from a generic `nanobind::object` even though it means converting the whole ndarray.

I'm thinking this same strategy may be simplify the whole "dispatch to the correct mangled c++ function name based on types" by instead always taking an untyped `nanobind::ndarray<>` and then on the c++ side doing the dispatch to the correct template instantiations. However... I'm not doing that now. Maybe some day.

.cscs-ci/container/deps.Containerfile
msimberg: https://github.com/spack/spack-packages/pull/4754 adds the nanobind dependency to the spack package.

.cscs-ci/default.yaml
```diff
@@ -4,7 +4,8 @@ include:
 variables:
   BASE_IMAGE: jfrog.svc.cscs.ch/docker-group-csstaff/alps-images/ngc-pytorch:26.01-py3-alps4-dev
   SPACK_SHA: v1.1.1
-  SPACK_PACKAGES_SHA: bc93746ce936d6653271b6e98f6df6ee28f64e84 # develop on 2026-03-25
+  # TODO: Use upstream commit
+  SPACK_PACKAGES_SHA: ghex-nanobind
```
msimberg:
```suggestion
  SPACK_PACKAGES_SHA: a010c65289743f900bdbbfb840e4d1876c24e93f # develop on 2025-05-08
```

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
```diff
@@ -21,47 +21,17 @@
...
-    else if constexpr (std::is_same_v<T, std::int64_t>)
-        return numpy_dtype(nanobind::str("int64"));
-    else
-        static_assert(sizeof(T) == 0, "unsupported dtype");
-}
-} // namespace
-
 std::string
-py_dtype_to_cpp_name(nanobind::handle dtype)
+py_dtype_to_cpp_name(nanobind::ndarray<> array)
```
philip-paul-mueller, line 27: I would add a note that this does not create a copy or switch to `const ndarray&` to make it clear.

philip-paul-mueller (approved):
LGTM

bindings/python/src/_pyghex/py_dtype_to_cpp_name.cpp
msimberg: https://github.com/ghex-org/GHEX/pull/198/commits/e15cb198ebc132e5a8116d1f6fd881866c739a87

PR #199, opened 2026-03-20
Invert meaning of operator bool for CUDA stream and event

msimberg:
This was a leftover from #190. See https://github.com/ghex-org/GHEX/pull/190#discussion_r2661802626.


msimberg:
I'll go ahead and merge this together with #200.

PR #200, opened 2026-03-23
Unify event implementation

msimberg:
This is on top of #199.

Uses the explicit implementation of a cudaEvent_t RAII wrapper in include/ghex.device/cuda/event.hpp in the device future implementation.

### DIFF
--- a/include/ghex/device/cuda/event.hpp
+++ b/include/ghex/device/cuda/event.hpp
@@ -27,7 +27,10 @@ struct cuda_event
     cudaEvent_t           m_event;
     ghex::util::moved_bit m_moved;
 
-    cuda_event() : cuda_event(cudaEventDisableTiming) {}
+    cuda_event()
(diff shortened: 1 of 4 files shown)


philip-paul-mueller (approved):
I have some comments/questions.

I kind of think that the number of places where we have amended the meaning is a bit low?

msimberg:
> I kind of think that the number of places where we have amended the meaning is a bit low?

@philip-paul-mueller do you mean inverting the bool operator meaning? I think it's because the `GHEX_C_MANAGED_STRUCT` wrapper actually had the `operator bool` semantics the correct way around from the beginning. So e.g. all the uses of `if (event)` in cuda/future.hpp were already the correct way around.

Or do you mean something else?

philip-paul-mueller:
No I meant exactly that, the number of places felt a bit low.
But considering the Macro it makes kind of sense.

PR #203, opened 2026-03-25
Update GPU architecture settings in pyproject.toml

havogt:


msimberg (approved):
Thanks, looks good to me. Should have no effect on uenv builds, but doesn't hurt to update these.

PR #209, opened 2026-05-13
Make sure the pyghex library explicitly links to oomph

msimberg:
The target already explicitly links to oomph::oomph, but oomph::oomph is an interface-only target. pyghex was not explicitly linking to the actual implementation library, i.e. oomph_mpi/libfabric/etc. It seems that somehow the library automatically found on linux, and on macos before #198. However, after #198 building the GHEX python package broke with unresolved symbols on macos (thanks @stubbiali for testing and pointing this out).

This PR simply explicitly links to the oomph_* library.


msimberg:
Confirmed working by @stubbiali. We really need a macos CI to catch this kind of change in the future.

stubbiali:
Thanks for the speedy patch @msimberg!

PR #207, opened 2026-04-30
Fix support for 1d regular domains

stubbiali:
Patch for #205.

PR #212, opened 2026-05-28
Fix `CartesianSet` util for 1d cases in python bindings

tehrengruber:
- In preparation of #207 (support for 1d regular-domain halo exchange) the `ProductSet` now supports 1d cases. IndexSpace becomes dimension-generic and decompose builds the per-rank subsets up front and uses the IndexSpace constructor.
The following changes are ported from PMAP where we use the same implementation. Eventually the helpers should be moved to a common package that GHEX, GT4Py and PMAP can use, instead of this awkward syncing, but this is beyond scope here:
- `ProductSet.without` and `UnionCartesian.without` are refactored to reduce recursion depth, which significantly improves performance.
- Rename `dim` to `ndim` to match the numpy convention, keeping `dim` as a deprecated alias that emits a DeprecationWarning.
- Adapt the callers to the new api: HaloGenerator builds its halo boxes without the removed ProductSet.from_coords, and the structured domain-descriptor test constructs IndexSpace with its subsets instead of the removed add_subset.

### DIFF
--- a/bindings/python/src/ghex/structured/cartesian_sets.py
+++ b/bindings/python/src/ghex/structured/cartesian_sets.py
@@ -8,21 +8,25 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 from __future__ import annotations
-from copy import copy
+
 import functools
+from functools import cached_property
 import itertools
 import math
 import operator
+import typing
+import warnings
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any, Dict, Literal, Sequence, Tuple, Union, TypeAlias
+    from collections.abc import Callable, Iterator, Sequence
+    from typing import Any, Literal, Optional, Union
 
-    integer: TypeAlias = Union[int, Literal[math.inf], Literal[-math.inf]]
+    Integer = Union[int, Literal[math.inf, -math.inf]]
 
 
-def is_integer_like(val):
-    return isinstance(val, int) or val == math.inf or val == -math.inf
+def is_integer_like(val: Any) -> bool:
+    return isinstance(val, int) or val in (math.inf, -math.inf)
 
 
 class Set:
@@ -32,34 +36,30 @@ class Set:
 class IntegerSet(Set):
     """A set containing integers."""
 
-    def empty_set(self):
+    @staticmethod
+    def empty_set() -> UnitRange:
         return UnitRange(0, 0)
 
-    def universe(self):
+    @staticmethod
+    def universe() -> UnitRange:
         return UnitRange(-math.inf, math.inf)
 
-    def shrink(self, arg: Union[int, Tuple[int, int]]):
-        if isinstance(arg, int):
-            arg = (arg, arg)
-
-        return self.extend(tuple(-v for v in arg))
-
     @staticmethod
-    def primitive_type():
+    def primitive_type() -> type[UnitRange]:
         return UnitRange
 
     @staticmethod
-    def union_type():
+    def union_type() -> type[UnionRange]:
         return UnionRange
 
 
 class UnitRange(IntegerSet):
     """Range from `start` to `stop` with step size one."""
 
-    start: integer
-    stop: integer
+    start: Integer
+    stop: Integer
 
-    def __init__(self, start: integer, stop: integer):
+    def __init__(self, start: Integer, stop: Integer) -> None:
         assert stop >= start
 
         self.start = start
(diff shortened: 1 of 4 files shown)


bindings/python/src/ghex/structured/cartesian_sets.py
```diff
@@ -8,21 +8,25 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 from __future__ import annotations
-from copy import copy
+
 import functools
+from functools import cached_property
```
stubbiali, line 13:
```suggestion
```

bindings/python/src/ghex/structured/cartesian_sets.py
```diff
@@ -70,30 +70,30 @@ def __init__(self, start: integer, stop: integer):
             self.start = 0
             self.stop = 0
 
-    @property
+    @cached_property
```
stubbiali, line 73:
```suggestion
    @functools.cached_property
```
This suggestion applies to all occurrences of `@cached_property`.

bindings/python/src/ghex/structured/cartesian_sets.py
```diff
@@ -133,32 +133,30 @@ def __getitem__(self, arg: Union[int, slice]):
 
         raise ValueError(f"Invalid argument `{arg}`")
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"UnitRange({self.start}, {self.stop})"
 
-    def __mul__(self, other: UnitRange):
+    def __mul__(self, other: UnitRange) -> Union[ProductSet, UnionRange]:
```
stubbiali, line 139:
```suggestion
    def __mul__(self, other: Set) -> Union[ProductSet, UnionCartesian, UnionRange]:
```

bindings/python/src/ghex/structured/cartesian_sets.py
```diff
@@ -272,69 +262,68 @@ def __init__(self, *args, disjoint=False):
...
+    def make_disjoint(self) -> UnionMixin:
         if self.disjoint:
             return self
 
         args = list(self.args)
-        for i, arg1 in enumerate(args):
-            for j, arg2 in enumerate(args[i + 1 :], start=i + 1):
-                args[j] = arg2.without(arg1, simplify=False)
+        for i in range(len(args)):
+            arg1 = args[i]
```
stubbiali, line 315:
```suggestion
        for i, arg1 in enumerate(args):
```

bindings/python/src/ghex/structured/cartesian_sets.py
```diff
@@ -366,191 +358,190 @@ def bounds(self) -> UnitRange:
...
-                    UnitRange(fused_args[-1].start, arg.stop),
-                ]
+                fused_args = [*fused_args[0:-1], UnitRange(fused_args[-1].start, arg.stop)]
             else:
                 fused_args.append(arg)
 
         return union(*fused_args, simplify=False, disjoint=self.disjoint)
 
-    def __mul__(self, other):
+    def __mul__(self, other) -> UnionCartesian:
```
stubbiali, line 373:
```suggestion
    def __mul__(self, other: Set) -> UnionCartesian:
```

bindings/python/src/ghex/structured/cartesian_sets.py
```diff
@@ -366,191 +358,190 @@ def bounds(self) -> UnitRange:
...
 
-    def __contains__(self, arg: Sequence[integer]):
+    def __contains__(self, arg: Sequence[Integer]) -> bool:
         assert all(is_integer_like(i) for i in arg)
         assert len(arg) == len(self.args)
 
         return all(i in r for r, i in zip(self.args, arg))
 
-    def issubset(self, arg):
+    def issubset(self, arg: Set):
```
stubbiali, line 544:
```suggestion
    def issubset(self, arg: Set) -> bool:
```

bindings/python/src/ghex/structured/cartesian_sets.py
```diff
@@ -649,63 +644,103 @@ def simplify(self):
...
+        return IndexSpace(new_subsets)
+
+    def intersect(self, mask: ProductSet) -> IndexSpace:
+        """Get a new :class:`IndexSpace` by masking all subsets."""
+        if self.ndim == 1:
+            m = ProductSet(mask.args[2])
+        elif self.ndim == 2:
+            m = mask.args[0] * mask.args[1]
+        else:
+            m = mask
```
stubbiali, line 688: This is a hack that we introduced in PMAP to deal with 2-d horizontal and 1-d vertical index spaces, but no longer need with https://github.com/PMAP-Project/PMAP/pull/332:
```suggestion
        assert mask.ndim == self.ndim
```

stubbiali (approved):
Thank you @tehrengruber, LGTM, I only have a few minor comments.

msimberg (approved):

PR #221, opened 2026-07-21
Add scheduled exchange to structured python bindings and test on stream

tehrengruber:
Share extract_cuda_stream via cuda_stream.hpp; parametrize the structured and unstructured exchange tests over gpu memory and stream (plain vs scheduled).

### DIFF
--- a/bindings/python/src/_pyghex/cuda_stream.hpp
+++ b/bindings/python/src/_pyghex/cuda_stream.hpp
@@ -1,0 +1,84 @@
+/*
+ * ghex-org
+ *
+ * Copyright (c) 2014-2023, ETH Zurich
+ * All rights reserved.
+ *
+ * Please, refer to the LICENSE file in the root directory.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#pragma once
+
+#include <ghex/config.hpp>
+
+#if defined(GHEX_CUDACC)
+
+#include <cstdint>
+#include <sstream>
+#include <type_traits>
+
+#include <ghex/device/cuda/runtime.hpp>
+
+#include <nanobind/nanobind.h>
+
+namespace pyghex
+{
+
+// Convert a Python stream object into a CUDA/HIP stream handle. The following
+// Python objects are understood:
+// - `None` is interpreted as the default stream, i.e. `nullptr`.
+// - an object implementing Nvidia's `__cuda_stream__` protocol, see
+//   https://nvidia.github.io/cuda-python/cuda-core/latest/interoperability.html#cuda-stream-protocol
+// - a CuPy stream, which exposes the stream handle via its `.ptr` attribute, see
+//   https://docs.cupy.dev/en/latest/reference/generated/cupy.cuda.Stream.html#cupy-cuda-stream
+inline cudaStream_t
+extract_cuda_stream(nanobind::object python_stream)
+{
+    static_assert(std::is_pointer<cudaStream_t>::value);
+    if (python_stream.is_none())
+    {
+        // NOTE: This is very C++ like, maybe remove and consider as an error?
+        return static_cast<cudaStream_t>(nullptr);
+    }
+    else
+    {
+        if (nanobind::hasattr(python_stream, "__cuda_stream__"))
+        {
+            nanobind::tuple cuda_stream_protocol =
+                nanobind::cast<nanobind::tuple>(python_stream.attr("__cuda_stream__")());
+            if (cuda_stream_protocol.size() != 2)
+            {
+                std::stringstream error;
+                error << "Expected a tuple of length 2, but got one with length "
+                      << cuda_stream_protocol.size();
+                throw nanobind::type_error(error.str().c_str());
+            }
+
+            const auto protocol_version = nanobind::cast<std::size_t>(cuda_stream_protocol[0]);
+            if (protocol_version != 0)
+            {
+                std::stringstream error;
+                error << "Expected `__cuda_stream__` protocol version 0, but got "
+                      << protocol_version;
+                throw nanobind::type_error(error.str().c_str());
+            }
+
+            const auto stream_address = nanobind::cast<std::uintptr_t>(cuda_stream_protocol[1]);
+            return reinterpret_cast<cudaStream_t>(stream_address);
+        }
+        else if (nanobind::hasattr(python_stream, "ptr"))
+        {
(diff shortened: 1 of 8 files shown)


msimberg:
@tehrengruber any plans for this one? Is it ready from your side? We could try to get this into the next release together with #222 and #224 if it's close to ready.

tehrengruber:
This is ready and by now also tested to be working in PMAP.

havogt (commented):
## Review

**No correctness bug in the C++ additions.** I checked and found sound:

- `extract_cuda_stream` moved verbatim to `pyghex` namespace scope as `inline`; both call sites resolve by enclosing-namespace lookup, the `<ghex/config.hpp>` guard is correct, and `src/_pyghex` is already on `target_include_directories`.
- The `communication_object_shim::schedule_exchange` overload pair (buffer_info pack vs. iterator pack) resolves the same way the existing `exchange` pair does; the index juggling in `schedule_exchange_from_iterators` produces `(b0, e0, b1, e1)` correctly.
- `has_scheduled_exchange() const` + `std::visit` over the `std::monostate`-fronted variant compiles and returns `false` for the empty state.
- `handle.wait()` after `schedule_wait()` does `cudaEventSynchronize` on the event recorded on the user stream, so the tests' subsequent `cp.asnumpy(...)` on the default stream is genuinely safe.
- The CUDA/HIP bindings builds in `.github/workflows/CI.yml` do compile the Python bindings, so the new `#if defined(GHEX_CUDACC)` blocks are compile-checked.

The substantive concern is that the PR *replaces* the async test rather than extending it, and net stream coverage goes down.

---

### 1. The new stream tests cannot detect a broken `schedule_wait`

`test/bindings/python/test_structured_pattern.py:138` (and `test_unstructured_domain_descriptor.py:288`)

Both new `exchange()` helpers call `handle.wait()` before reading any result, and `wait()` funnels into `complete_schedule_exchange()` → host-blocking `cudaEventSynchronize`. That host sync alone makes the results visible, so both tests would still pass if `schedule_sync_unpack()` no longer ordered the unpack kernels against the user stream — i.e. the stream-ordering guarantee that is the entire point of `schedule_exchange`/`schedule_wait` is never exercised.

The deleted `test_domain_descriptor_async` did exercise it: `check_field(d1, "C", stream)` between `schedule_wait()` and `wait()`, with `cp.asnumpy(..., stream=stream, blocking=True)`. Worth keeping a check that reads back synchronizing *only* on the scheduled stream, before `wait()`.

### 2. The `__cuda_stream__` and null-stream branches of `extract_cuda_stream` lose their only tests

`test/bindings/python/test_unstructured_domain_descriptor.py:13`

`CUDAStreamProtocolMock` and `STREAM_TYPES_TO_TEST` are deleted, and the new parametrization only ever constructs `cp.cuda.Stream`, which is picked up by the `.ptr` fallback. cupy's `Stream` has no `__cuda_stream__` attribute — that is exactly why the mock existed — so the protocol branch (tuple-length check, version check, `reinterpret_cast` of element 1) is now dead code as far as the suite is concerned.

The `stream=None` → `nullptr` path also loses its test: `stream_args is None` routes to the plain `co.exchange()` path, so `schedule_exchange(None, ...)` is never called. `{"null": True}` reaches the same handle through `.ptr`, not through `is_none()`.

Given that this PR's stated purpose is to promote `extract_cuda_stream` into a shared header used by two binding modules, it widens the blast radius while shrinking the tests.

### 3. Leaked cartesian communicator, now 4× more often

`test/bindings/python/test_structured_pattern.py:80`

`mpi_comm.Create_cart(...)` is never `Free()`d here (pre-existing), but the parametrization grows from 6 to 24 invocations. `test/bindings/python/fixtures/context.py:50-53` documents explicitly that failing to free these communicators causes subsequent parallel tests to hang with the UCX backend. Since the PR quadruples the leak in the one test that rolls its own communicator, adding the `Free()` — or reusing the `mpi_cart_comm` fixture — is cheap insurance.

### 4. `has_scheduled_exchange` is registered inside the inner per-field loop

`bindings/python/src/_pyghex/structured/regular/communication_object.cpp:124`

Unlike unstructured (where `_communication_object` is a fresh class per co-specialization), the structured `_communication_object` is a single `register_class<communication_object_shim>` created once at line 35, while the inner `for_each` runs for every (co-specialization × field-descriptor) pair. The lambda `[](communication_object_shim&) -> bool` has a signature independent of `buffer_info_type`, so the same byte-identical overload gets registered ~270 times (`types::data` 5 × `types::archs` 2 × layouts 1+2+6+18), all but the first unreachable. Cost is binary size and module-init time. Move the `.def("has_scheduled_exchange", …)` next to line 35, outside both loops.

### 5. The new GPU/stream parametrizations never run under `ctest`

`bindings/python/min-requirements-test.txt`

`cupy` is not in the test requirements, so in the venv built by `test/bindings/python/CMakeLists.txt` all three `(True, …)` variants hit `pytest.skip("`CuPy` is not installed.")` even on a CUDA-enabled build. The added coverage only materialises for someone running pytest in a hand-made environment. Worth either adding an optional cupy install for GPU configs, or noting explicitly that these are manual-only.

### 6. Nit: `cp.asnumpy(field)` omits `order=` that the sibling test claims is required

`test/bindings/python/test_structured_pattern.py:161`

Fields are created with `order="F"` (line 112) but converted back with the default `order="C"`, while `test_unstructured_domain_descriptor.py:272` in this same PR keeps `cp.asnumpy(data, order=order)` with the note *"Without the explicit order it fails sometimes."*

`asnumpy` does a logical copy, so element values are preserved either way and this is almost certainly not a real flake source — which suggests the comment in the unstructured test is misleading and should go. Either way the two files should not disagree about it.

### 7. Nit: unused import

`test/bindings/python/test_unstructured_domain_descriptor.py:18` — `from ghex.context import make_context` is added but never used; the test gets its context from the `cart_context` fixture. (`capsys` at line 224 is also unused, but that is pre-existing.)

---

### Outside the diff

`bindings/python/src/_pyghex/structured/types.hpp:51` — `layouts<4>::list` enumerates only 18 of the 24 permutations: every `layout_map<2, …>` is missing (the `0…`, `1…` and `3…` groups are present, the `2…` group is not). A 4-D field whose first array dimension has the third-largest stride fails in `cls_from_cpp_type_spec` with a mangled-name lookup error. Pre-existing, but this PR touches the registration code that consumes the list.

msimberg (approved):

PR #222, opened 2026-08-04
Fix unstructured GPU exchange silently corrupting halos with more than blockDim.y * 65535 indices

havogt:
The unstructured GPU pack/unpack kernels (`unstructured/user_concepts.hpp`) put `ceil(num_indices / blockDim.y)` into `gridDim.y`, which CUDA limits to 65535. With the current block size (`GHEX_UNSTRUCTURED_SERIALIZATION_THREADS_PER_BLOCK_Y = 8`), any halo with more than 8 * 65535 = 524280 indices per neighbour makes the kernel launch fail with an invalid configuration. Since the launch error is never checked, the exchange completes without any error and the halo is simply left untouched.

Observed on santis (GH200): a 2-rank unstructured exchange with 1048576 halo indices per rank leaves the entire halo at its pre-exchange values, on both the send side (pack) and the receive side (unpack), for both cray-mpich 8.x and 9.x — the bug is in the kernel launch, not the transport.

This PR adds a failing test (`test/unstructured/test_large_halo.cpp`, runs as `parallel-ranks-2` on the GPU CI): a quiescent 2-rank exchange with 600000 halo indices per neighbour, verifying the halo contents afterwards. CI is expected to be red on the GPU pipelines.

The fix is now included: the levels-first kernels map the index blocks to gridDim.x (limit 2^31-1) instead of gridDim.y, keeping the thread mapping — and thus memory coalescing — unchanged, and all four pack/unpack launches are followed by a launch-error check so an invalid configuration can never pass silently again. The test additionally covers more levels than fit in one block (exercising the level-block grid dimension) and derives the overflowing halo size from the block-size macro. Validated on santis: the test is green after the fix, and the repeated-exchange hang described above no longer occurs. The red state is preserved in the CI runs of the earlier, test-only pushes.

---

Update: the failure mode is worse than silent corruption. In a loop of repeated exchanges with an oversized halo, the first exchange completes (with the halo untouched) and a subsequent exchange hangs until the job time limit (observed on santis with both cray-mpich 8.x and 9.x, intra- and inter-node). The single-exchange test in this PR does not hang, it fails the content check.

### DIFF
--- a/test/unstructured/test_large_halo.cpp
+++ b/test/unstructured/test_large_halo.cpp
@@ -1,0 +1,97 @@
+/*
+ * ghex-org
+ *
+ * Copyright (c) 2014-2026, ETH Zurich
+ * All rights reserved.
+ *
+ * Please, refer to the LICENSE file in the root directory.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// The unstructured GPU pack/unpack kernels place ceil(num_indices / blockDim.y)
+// in gridDim.y, which CUDA limits to 65535. A halo with more than
+// blockDim.y * 65535 indices per neighbour makes the kernel launch fail with an
+// invalid configuration; the missing launch-error check lets the exchange
+// complete silently with the halo left untouched.
+
+#include <gtest/gtest.h>
+#include "../mpi_runner/mpi_test_fixture.hpp"
+
+#include <ghex/config.hpp>
+#include <ghex/unstructured/pattern.hpp>
+#include <ghex/unstructured/user_concepts.hpp>
+#include <ghex/communication_object.hpp>
+#include "../util/memory.hpp"
+
+#include <numeric>
+#include <set>
+#include <vector>
+
+using domain_id_type = int;
+using global_index_type = int;
+using domain_descriptor_type =
+    ghex::unstructured::domain_descriptor<domain_id_type, global_index_type>;
+using halo_generator_type = ghex::unstructured::halo_generator<domain_id_type, global_index_type>;
+using grid_type = ghex::unstructured::grid;
+
+#ifdef GHEX_CUDACC
+
+using data_descriptor_gpu_type =
+    ghex::unstructured::data_descriptor<ghex::gpu, domain_id_type, global_index_type, int>;
+
+namespace
+{
+// larger than GHEX_UNSTRUCTURED_SERIALIZATION_THREADS_PER_BLOCK_Y * 65535
+constexpr int halo_size = 600000;
+constexpr int inner_size = halo_size;
+
+// Each rank owns gids [rank*inner_size, (rank+1)*inner_size); its halo is the
+// first halo_size gids of the other rank, stored after the inner elements.
+domain_descriptor_type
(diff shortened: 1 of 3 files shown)


havogt:
@msimberg this is not **the** bug. Will post another one soon.

havogt:
cscs-ci run default

msimberg:
cscs-ci run

test/unstructured/test_large_halo.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#ifdef GHEX_CUDACC
+
+using data_descriptor_gpu_type =
+    ghex::unstructured::data_descriptor<ghex::gpu, domain_id_type, global_index_type, int>;
+
+namespace
+{
+// larger than GHEX_UNSTRUCTURED_SERIALIZATION_THREADS_PER_BLOCK_Y * 65535
+constexpr int halo_size = 600000;
```
msimberg: This should possibly just compute a value large enough based on GHEX_UNSTRUCTURED_SERIALIZATION_THREADS_PER_BLOCK_Y instead (in case it changes)?

msimberg (commented):
Thank you @havogt/claude. Minor comment on the test (it seems to be failing correctly in ci, good!). As for the fix, I think loudly failing on launch failures would be a sufficient fix at the moment (especially since we should anyway be checking for other launch failures, that in itself is a bug). If the launch bounds can be deteceted at runtime chunking the packing/unpacking dynamically could also be an option, but I think it may not be necessary for now... as long as the failure is loud.

havogt:
Some measurements on the performance impact of this change (GH200, CUDA 13.1, standalone microbenchmark of the levels-first pack kernel, event-timed over 2000 launches; `old` = index blocks in gridDim.y, `new` = this PR):

| case | old | new | grid-stride alternative |
|---|---|---|---|
| 500k indices × 80 levels (160 MB) | 237.5 µs | +2.6% | +3.8% |
| 600k indices × 1 level | 0.85 µs — silent no-op (the bug) | 48.6 µs (works) | 50.0 µs |
| 50k × 80 | 15.1–15.4 µs | −0.1% … +2.3% (run-to-run noise) | +8% |
| 20k × 60 | 5.5 µs | ±0.2% | +7% |
| 4k × 1 | 2.4 µs | ±0.1% | +0–4% |

- The block-order swap is free at realistic halo sizes; the only stable difference is +2.6% kernel-only in the narrow window just below the old 524280-index limit, where the exchange is dominated by transferring a ≥150 MB message anyway. Above the limit the old kernel "wins" only by not running at all.
- A grid-stride variant (keeping the old block order, capping gridDim.y) was measured as well and is strictly worse (+4-8% at mid sizes), so the simple swap was kept.
- `cudaGetLastError()` after a launch costs 9.5 ns on this system (it only reads a thread-local error slot, no device synchronization) next to a ~1.7 µs kernel launch, i.e. not measurable. It is the call that reports invalid launch configurations immediately.

havogt:
cscs-ci run

msimberg (approved):

PR #223, opened 2026-08-04
GPU unpack from inline receive completion races with pre-exchange kernels on the user stream

havogt:
The oomph MPI backend completes receives inline when `MPI_Test` succeeds directly after `MPI_Irecv` (`src/mpi/communicator.hpp`). For GPU fields the completion callback launches the unpack kernel on the buffer's non-blocking stream. This happens inside `post_recvs()`, i.e. before `post_sends()` has drained the device, and nothing orders the unpack kernel after work submitted to the user stream before `schedule_exchange()`. Kernels still running on the user stream can therefore read halo elements while the unpack overwrites them. The deferred-callback path is safe only incidentally: `post_sends()` host-blocks on pack kernels which wait on the user-stream event recorded in `schedule_sync_pack()`.

This was originally observed as sporadic verification failures in ICON + ICON4Py on Alps after a cray-mpich upgrade; the exchanged field itself always ends up correct, only consumers of the halo that were already running are affected — which made it look like a compiler/codegen bug for a long time.

The test creates rank imbalance (rank 1 runs a long halo-reading kernel on the user stream and delays posting its receives while progressing MPI) so that the neighbour's small message has already arrived at `MPI_Irecv` time. Per-pass halo checksums in the reading kernel detect any unpack overlap.

Measured on santis (GH200, 2 ranks on 2 nodes): which message size fires varies from run to run (likely with node placement), so each invocation covers four eager-sized messages (512 B, 2 KB, 16 KB, 32 KB) with 12 skewed trials per size and the skew staggered across trials. With these default settings and cray-mpich 9.1.0, 5 out of 5 test invocations failed without the fix — notably via a different message size each time — and 5 out of 5 passed with it (torn halo reads were observed twice across the campaign). With cray-mpich 8.x the branch fires less often but overlap was still observed in every full sweep. Intra-node the branch never fired on either version. One invocation takes ~85 s, hence the ctest timeout increase for parallel tests (which recently also hit spurious timeouts elsewhere). Deferring all callbacks (forcing the else-branch in oomph) eliminates the overlap completely. Whether the CI transport (external OpenMPI 5.0.10) completes receives at post time is exactly what this PR should reveal — hence the new 2-node CI jobs; the single-node jobs are not expected to trigger it.

Note the test is expected to be red *probabilistically* on affected transports. A fix (ordering the unpack streams after the user stream before `post_recvs()`) follows in a separate PR, which makes the test deterministically green.

Env knobs for manual investigation: `GHEX_TEST_INLINE_UNPACK_RACE_HALO` (ints per halo), `..._PASSES`, `..._SPIN_CYCLES`, `..._SLEEP_MS`, `..._CONTROL_TRIALS`, `..._TRIALS`.

### DIFF
--- a/.cscs-ci/default.yaml
+++ b/.cscs-ci/default.yaml
@@ -95,7 +95,7 @@ build_libfabric:
   extends: .container-runner-clariden-gh200
   variables:
     SLURM_GPUS_PER_TASK: 1
-    SLURM_TIMELIMIT: '5:00'
+    SLURM_TIMELIMIT: '10:00'
     SLURM_PARTITION: normal
     SLURM_MPI_TYPE: pmix
     SLURM_NETWORK: disable_rdzv_get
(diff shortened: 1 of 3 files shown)


msimberg:
cscs-ci run

havogt:
cscs-ci run

havogt:
cscs-ci run

havogt:
cscs-ci run

msimberg:
I'll close this, let's merge it all in #224.

PR #224, opened 2026-08-04
Fix: order unpack streams after the user stream in schedule_exchange

havogt:
Fix for the race demonstrated in #223 (stacked on that branch; review only the last commit).

`post_recvs()` may invoke the receive-completion callback inline when the transport completes the receive at posting time, which for GPU fields launches the unpack kernel on the buffer's non-blocking stream — with no ordering against work submitted to the user stream before `schedule_exchange()`. Kernels still running on the user stream can read halo elements mid-overwrite.

The fix extends the existing event dependency in `schedule_sync_pack()` (renamed to `schedule_sync_pack_unpack()`) to the receive-buffer streams: the event recorded on the user stream now also gates every unpack stream. An inline-launched unpack kernel is still enqueued immediately — inline completion keeps its latency/imbalance benefit of overlapping unpack with packing and sending — but executes only after previously submitted user-stream work has completed, which is the earliest legal point anyway.

The plain `exchange()` overload takes no stream to order against; its implicit contract (device work on the exchanged fields must be complete when called) is now documented.

Validated on santis (GH200, 2 nodes, cray-mpich 8.x and 9.1.0): with this change the #223 test (at its default settings) passes in 5 out of 5 invocations with 0 overlapping trials while the inline branch still fires; without it all 5 invocations fail.

### DIFF
--- a/.cscs-ci/default.yaml
+++ b/.cscs-ci/default.yaml
@@ -95,7 +95,7 @@ build_libfabric:
   extends: .container-runner-clariden-gh200
   variables:
     SLURM_GPUS_PER_TASK: 1
-    SLURM_TIMELIMIT: '5:00'
+    SLURM_TIMELIMIT: '10:00'
     SLURM_PARTITION: normal
     SLURM_MPI_TYPE: pmix
     SLURM_NETWORK: disable_rdzv_get
@@ -121,7 +121,7 @@ build_libfabric:
     # writing inside the container.
     - if [[ "${SLURM_LOCALID}" == 0 ]]; then rm -rf /ghex/build/Testing; mkdir /tmp/Testing; ln -s /tmp/Testing /ghex/build/Testing; fi
     - until [[ -L /ghex/build/Testing ]]; do sleep 1; done
-    - spack -e ci build-env ghex -- ctest --test-dir /ghex/build -L "parallel-ranks-${SLURM_NTASKS}" --output-on-failure --timeout 60
+    - spack -e ci build-env ghex -- ctest --test-dir /ghex/build -L "parallel-ranks-${SLURM_NTASKS}" --output-on-failure --timeout 180
 
 .test_parallel_job:
   extends: .test_parallel_template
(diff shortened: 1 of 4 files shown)


msimberg:
cscs-ci run

havogt:
cscs-ci run

test/unstructured/test_inline_unpack_race.cpp
```diff
@@ -0,0 +1,289 @@
...
+    return trials_with_overlap;
+}
+} // namespace
+
+TEST_F(mpi_test_fixture, inline_unpack_race)
+{
+    if (world_size != 2) GTEST_SKIP() << "test requires exactly 2 ranks";
+
+    case_params p;
+    p.num_passes = static_cast<int>(env_or("RACE_PASSES", 4000));
```
msimberg: I think these should be
```suggestion
    p.num_passes = static_cast<int>(env_or("GHEX_TEST_INLINE_UNPACK_RACE_PASSES", 4000));
```
etc.

.cscs-ci/default.yaml
```diff
@@ -185,6 +185,12 @@ test_parallel_2_mpi:
   variables:
     SLURM_NTASKS: 2
 
+test_parallel_2_mpi_2nodes:
+  extends: .test_parallel_mpi
+  variables:
+    SLURM_NTASKS: 2
+    SLURM_NTASKS_PER_NODE: 1
+
```
msimberg, line 205: I think rerunning all backends x 2/4 nodes in addition to the current single-node tests is a cheap enough addition that we can just always run them. Do you think you could add the other combinations as well?

havogt:
cscs-ci run

test/unstructured/test_inline_unpack_race.cpp
```diff
@@ -0,0 +1,291 @@
...
+reader_kernel(const int* field, int offset, int n, long long spin_cycles, int num_passes,
+    unsigned long long* sums)
+{
+    for (int p = 0; p < num_passes; ++p)
+    {
+        unsigned long long s = 0;
+        for (int i = 0; i < n; ++i) s += static_cast<const volatile int*>(field)[offset + i];
+        sums[p] = s;
+        const long long start = clock64();
+        while (clock64() - start < spin_cycles) {}
```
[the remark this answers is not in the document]
havogt: `clock64()` is signed: the CUDA docs declare it as `long long int clock64()`, so the code matches the documented signature and `clock64() - start` with `long long` is the standard spin idiom. Overflow would require 2^63 cycles, i.e. on the order of a century of GPU uptime. Leaving as is.

msimberg (approved):

PR #226, opened 2026-08-05
Register has_scheduled_exchange once per communication object

havogt:
`has_scheduled_exchange` is `.def`'d on `_communication_object` from inside the `for_each` over `field_descriptor_specializations`, but unlike the `exchange`/`schedule_exchange` bindings next to it, its lambda signature (`[](type&) -> bool`) does not depend on `buffer_info_type`. The identical overload is therefore registered once per field specialization — `types::data` (5) x `types::archs` (2) = ten times per communication-object specialization — and every registration after the first is unreachable.

No behaviour change; it removes dead overloads from the module-init path.

Found while reviewing #221, which adds the same binding to the structured bindings (with a much larger multiplier there, since the structured `_communication_object` is a single class registered outside both loops).


havogt:
cscs-ci run

msimberg (approved):

